PR #24, opened 2016-01-28
Create gateway behaviors directly in feature setup

janovesk:
Reduce the dependency injection footprint.

### DIFF
--- a/src/NServiceBus.Gateway/Gateway.cs
+++ b/src/NServiceBus.Gateway/Gateway.cs
@@ -38,9 +38,6 @@
         /// </summary>
         protected override void Setup(FeatureConfigurationContext context)
         {
-            context.Pipeline.Register<GatewayIncomingBehavior.Registration>();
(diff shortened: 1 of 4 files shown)


janovesk:
Is this what you had in mind @danielmarbach?

src/NServiceBus.Gateway/Gateway.cs
```diff
@@ -59,18 +56,15 @@ protected override void Setup(FeatureConfigurationContext context)
             ConfigureChannels(context);
 
             ConfigureReceiver(context);
-            ConfigureSender(context, gatewayPipeline, gatewayInputAddress);
-            ConfigureRouteToGateway(context, gatewayInputAddress);
+            context.Pipeline.Register("GatewayIncomingBehavior", b => new GatewayIncomingBehavior(), "Extracts gateway related information from the incoming message");
```
danielmarbach: You can use the overload which doesn't provide the builder here.

danielmarbach:
Yes!

PR #28, opened 2016-02-08
Refactored the HttpChannelReceiver to use same approach as the message pump in core

janovesk:
ping @danielmarbach

### DIFF
--- a/src/NServiceBus.Gateway/Channels/Http/HttpChannelReceiver.cs
+++ b/src/NServiceBus.Gateway/Channels/Http/HttpChannelReceiver.cs
@@ -19,8 +19,11 @@ namespace NServiceBus.Gateway.Channels.Http
         {
             dataReceivedHandler = dataReceivedOnChannel;
 
-            concurencyLimiter = new SemaphoreSlim(maxConcurrency, maxConcurrency);
-            tokenSource = new CancellationTokenSource();
+            concurrencyLimiter = new SemaphoreSlim(maxConcurrency, maxConcurrency);
+
+            cancellationTokenSource = new CancellationTokenSource();
+            cancellationToken = cancellationTokenSource.Token;
+
             listener = new HttpListener();
 
             listener.Prefixes.Add(address);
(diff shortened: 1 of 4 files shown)


src/NServiceBus.Gateway/Receiving/SingleCallChannelReceiver.cs
```diff
@@ -39,7 +39,7 @@ public void Dispose()
 
         void DisposeManaged()
         {
-            channelReceiver?.Dispose();
+            channelReceiver?.Stop();
```
danielmarbach: I think this will never be called since the outer class doesn't implement `IDisposable`. Should you just stop the `SingleCallChannelReceiver` which then stops the inner receiver?

src/NServiceBus.Gateway/Channels/Http/HttpChannelReceiver.cs
```diff
@@ -35,54 +38,55 @@ public void Start(string address, int maxConcurrency, Func<DataReceivedOnChannel
...
-                    var worker = Task.Run(() => Handle(context, cancellationToken), CancellationToken.None);
-                    runningTasks.TryAdd(worker, worker);
+                    if (cancellationTokenSource.IsCancellationRequested)
+                    {
+                        return;
+                    }
+
+                    await concurrencyLimiter.WaitAsync(cancellationToken).ConfigureAwait(false);
+
+                    var receiveTask = Task.Run(() => HandleMessage(context, cancellationToken), CancellationToken.None);
```
danielmarbach: `HandleMessage` is pure IO, right? If yes then remove `Task.Run`

src/NServiceBus.Gateway/Receiving/SingleCallChannelReceiver.cs
```diff
@@ -31,15 +31,9 @@ public void Start(Channel channel, int maxConcurrency, Func<MessageReceivedOnCha
...
-            //Injected at compile time
-            DisposeManaged();
-        }
-
-        void DisposeManaged()
-        {
-            channelReceiver?.Dispose();
+        public Task Stop()
+        { 
+            return channelReceiver?.Stop();
```
danielmarbach: Elvis doesn't work with `await`. I have no idea if it works here. Can you verify?

danielmarbach: @janovesk just showed me that elvis works in this scenario. It doesn't work with `await`. So

   await c?.Foo();

would throw a `NullReferenceException`

PR #35, opened 2016-02-19
[WIP] Readded MSMQ correlation id backwards compatibility hack

janovesk:
This got broken as part of the V6 rewrite. Added it back.


janovesk:
Turns out this code has never worked in version 1 gateway/version 5 core. `IsMsmqTransport` was never configured. Not fixing it now, since the code was marked [obsolete](https://github.com/Particular/NServiceBus/commit/a94795497153a6152e78f34061c548b5c19b0d61) from core version 6 anyway.

PR #36, opened 2016-02-19
Removed v6 obsoletes

janovesk:
This code was marked as [obsolete for version 6](https://github.com/Particular/NServiceBus/commit/a94795497153a6152e78f34061c548b5c19b0d61), but this got lost in translation when the code was moved from [core](https://github.com/Particular/NServiceBus/blob/46f5a2acee424a94a62fbd5645e093635c09cdd9/src/NServiceBus.Core/Gateway/HeaderManagement/HeaderMapper.cs) to its own [gateway package](https://github.com/Particular/NServiceBus.Gateway/blob/18dbba91993c878ec80b2c9a32d46415cf7638bd/src/NServiceBus.Gateway/HeaderManagement/HeaderMapper.cs).

The code never worked in version 5 because of missing initialization of the `IsMsmqProperty`, but [not fixing it](https://github.com/Particular/NServiceBus.Gateway/pull/35) as it is obsolete now anyway.

Removing it all.

### DIFF
--- a/src/NServiceBus.Gateway/Receiving/SingleCallChannelReceiver.cs
+++ b/src/NServiceBus.Gateway/Receiving/SingleCallChannelReceiver.cs
@@ -88,11 +88,6 @@
                 {
                     args.Headers = MapCustomMessageHeaders(headers);
                 }
-                
-                if (IsMsmqTransport)
-                {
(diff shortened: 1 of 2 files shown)


janovesk:
@andreasohlund take a look at this. You removed the obsoletes when moving the code from core to the separate gateway package. 

That was not intentional? 

We can remove this safely in core version 6/gateway version 2? If so, push the green button. :)

PR #38, opened 2016-02-24
update to core 1587

SimonCropp:
connects to #39

### DIFF
--- a/src/NServiceBus.Gateway.AcceptanceTests/NServiceBus.Gateway.AcceptanceTests.csproj
+++ b/src/NServiceBus.Gateway.AcceptanceTests/NServiceBus.Gateway.AcceptanceTests.csproj
@@ -32,15 +32,15 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="NServiceBus.AcceptanceTesting, Version=6.0.0.0, Culture=neutral, PublicKeyToken=9fc386479f8a226c, processorArchitecture=MSIL">
-      <HintPath>..\packages\NServiceBus.AcceptanceTesting.6.0.0-unstable1509\lib\net452\NServiceBus.AcceptanceTesting.dll</HintPath>
+      <HintPath>..\packages\NServiceBus.AcceptanceTesting.6.0.0-unstable1587\lib\net452\NServiceBus.AcceptanceTesting.dll</HintPath>
       <Private>True</Private>
     </Reference>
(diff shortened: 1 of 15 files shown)


SimonCropp:
@SzymonPobiega can u review my routing changes here

SzymonPobiega:
I see that the biggest difference is the fact that the callbacks package now requires the instance ID to be present and because it is `EnabledByDefault`, all the tests need to specify it.

How about, instead of using the callbacks package, you simulate this? [Here](https://github.com/Particular/NServiceBus.RabbitMQ/commit/302f2848b69d4fc6728c4fbe75b6b12f9a2a3057#diff-10ce60139aa573b172adda22f364c711R54) I've did similar change in Rabbit. The "trick" is to use `sendOptions.RouteReplyToThisInstance();` when sending a request (this is the only thing callbacks do in V6).

adamralph:
@SimonCropp is there any issue which we can connect this with?

SimonCropp:
@SzymonPobiega the callbacks is only required by the acceptance tests to verify they actually work. not required by the gateway nuspec https://github.com/Particular/NServiceBus.Gateway/blob/develop/packaging/nuget/nservicebus.gateway.nuspec or referenced in the gateway dll  https://github.com/Particular/NServiceBus.Gateway/blob/develop/src/NServiceBus.Gateway/NServiceBus.Gateway.csproj#L46


with that in mind does your comment still stand?

SzymonPobiega:
@SimonCropp yes. The problem with callbacks package I see is that it requires setting up instance ID in each and every endpoint in the acceptance tests. These IDs are not required for gateway so it blurs the tests. 

An alternative would be excluding the callbacks dll from scanning in the test infrastructure and include it only in that one test that refers to callbacks.

SzymonPobiega:
Let me try to validate that approach.

SimonCropp:
@SzymonPobiega ok cool. feel free to push directly to this branch/pr

SzymonPobiega:
OK, I give up. The gateway in its current shape is broken beyond repair... One thing only: the tests `When_doing_request_reply` and `When_doing_request_reply_between_sites` seem to be almost identical. Not sure if we need both.

danielmarbach:
ping @janovesk

janovesk:
> OK, I give up. The gateway in its current shape is broken beyond repair...

That's a big statement. Can you be a bit more explicit: what is broken? @SzymonPobiega

SzymonPobiega:
@janovesk that was a moment of frustraction. I meant, I cannot be repaired in the sense of having more than one URL per site.

timbussmann:
so is this one ready for merging? @SzymonPobiega @SimonCropp

SzymonPobiega:
:+1:

PR #45, opened 2016-04-14
add api approval

SimonCropp:


SimonCropp:
note there is a class `NServiceBus.BusExtensions` which should be renamed since it is not a "bus" anymore. also it is fairly generic name to have in the root namespace

PR #47, opened 2016-04-26
Fixed ChannelFactories signature

boblangley:
The receive channel func was set to return an `IChannelSender` instead of `IChannelReceiver`


SimonCropp:
suspect u will need to re-approve the api

boblangley:
Fixed API signature. Thanks @SimonCropp

bording:
:+1: Looks good and a small enough change that I'll go ahead and merge it since it's all green.

Issue #48, opened 2016-04-26
Add acceptance tests for custom channel factories


boblangley:
We do not have any test coverage for this functionality. This was discovered while creating snippets for doco and it was discovered the receive func had the wrong signature, #47 was opened to correct the signature.

We may also want to [create samples](https://github.com/Particular/docs.particular.net/issues/1458) as well.

PR #54, opened 2016-07-29
Reimplement gateway recoverability

MarcinHoppe:
Connects to Particular/V6Launch#69


tmasternak:
LGTM

danielmarbach:
Doesn't seem to build right now

MarcinHoppe:
@danielmarbach It's still WIP, does not include yesterday changes in the core.

tmasternak:
> LGTM

What I meant by that is: From recoverabilty/routing api perspective it looks good.

MarcinHoppe:
@Particular/nservicebus-maintainers This is now ready for review.

src/NServiceBus.Gateway.AcceptanceTests/When_sending_fails_with_retries.cs
```diff
@@ -163,5 +170,38 @@ public Task Handle(AnyMessage errorMessage, IMessageHandlerContext context)
                 Context testContext;
             }
         }
+
+        class FaultyChannelSender : IChannelSender
```
danielmarbach: Would it make sense to extract those impls to a root folder and reuse it in as well ?

https://github.com/Particular/NServiceBus.Gateway/pull/54/files#diff-8910b6948833c6ad5dcd6fd9e87e7e9eR120

src/NServiceBus.Gateway/Gateway.cs
```diff
@@ -50,28 +50,47 @@ protected override void Setup(FeatureConfigurationContext context)
...
+            var requiredTransactionSupport = context.Settings.GetRequiredTransactionModeForReceives();
 
-            gatewayPipeline.Register(new RetryFailedGatewayMessagesBehavior.Registration(gatewayInputAddress, retryPolicy));
-            gatewayPipeline.Register("GatewaySendProcessor", b => new GatewaySendBehavior(gatewayInputAddress, channelManager, new MessageNotifier(), b.Build<IDispatchMessages>(), context.Settings, CreateForwarder(channelSenderFactory, b.BuildAll<IDataBus>()?.FirstOrDefault()), GetConfigurationBasedSiteRouter(context)), "Processes messages to be sent to the gateway");
-            context.Pipeline.Register("RouteToGateway", b => new RouteToGatewayBehavior(gatewayInputAddress), "Reroutes gateway messages to the gateway");
+            var retryPolicy = context.Settings.HasSetting("Gateway.Retries.RetryPolicy") ? context.Settings.Get<Func<IncomingMessage, Exception, int, TimeSpan>>("Gateway.Retries.RetryPolicy") : DefaultRetryPolicy.BuildWithDefaults();
```
danielmarbach: Would it make sense to set the default policy as default in the feature ctor? Then you could always get the policy

src/NServiceBus.Gateway/Gateway.cs
```diff
@@ -50,28 +50,47 @@ protected override void Setup(FeatureConfigurationContext context)
...
 
+            context.AddSatelliteReceiver("Gateway", gatewayInputAddress, requiredTransactionSupport, PushRuntimeSettings.Default,
+                (config, errorContext) =>
+                {
+                    var currentRetry = errorContext.DelayedDeliveriesPerformed + 1;
```
danielmarbach: Would it make sense to extract this into static class and have a few tests around it?

src/NServiceBus.Gateway/Gateway.cs
```diff
@@ -139,9 +160,9 @@ static ConfigurationBasedSiteRouter GetConfigurationBasedSiteRouter(FeatureConfi
             return new ConfigurationBasedSiteRouter(sites);
         }
 
-        static void RegisterHttpListenerInstaller(FeatureConfigurationContext context, IManageReceiveChannels channelManager)
+        static void RegisterHttpListenerInstaller(FeatureConfigurationContext context, IManageReceiveChannels channelManager, bool enableHttpListener)
```
danielmarbach: Forgive my ignorance: What does this flag do?

danielmarbach:
@MarcinHoppe first round of review done

src/NServiceBus.Gateway/Gateway.cs
MarcinHoppe: @danielmarbach This flag allows us to disable the HTTP listener when custom channels are used. Previous implementation would not configure the `INeedToInstallSomething` implementation that started the listener when custom channels were used, which would lead to endpoint startup failure. 

I assume this would be done as part of #48 but I needed it earlier.

Does this make sense?

danielmarbach: makes sense

src/NServiceBus.Gateway.AcceptanceTests/When_sending_fails_with_retries.cs
MarcinHoppe: Fixed.

src/NServiceBus.Gateway/Gateway.cs
MarcinHoppe: Makes sense. Fixed.

src/NServiceBus.Gateway/Gateway.cs
MarcinHoppe: Makes sense. Fixed.

MarcinHoppe:
@danielmarbach Ready for a second round :).

danielmarbach:
@marcinhoppe have you been able to donthe smoke tests?

src/NServiceBus.Gateway.Tests/APIApprovals.Approve.approved.txt
```diff
@@ -13,7 +13,7 @@ namespace NServiceBus
     public class GatewaySettings
     {
         public void ChannelFactories(System.Func<string, NServiceBus.Gateway.IChannelSender> senderFactory, System.Func<string, NServiceBus.Gateway.IChannelReceiver> receiverFactory) { }
-        public void CustomRetryPolicy(System.Func<NServiceBus.Transports.IncomingMessage, System.Exception, int, System.TimeSpan> customRetryPolicy) { }
+        public void CustomRetryPolicy(System.Func<NServiceBus.Transport.IncomingMessage, System.Exception, int, System.TimeSpan> customRetryPolicy) { }
```
danielmarbach: Sorry didn't catch this in the first round. Should we call this CustomDelayedRetryPolicy since we only allow to customize the delayed part

src/NServiceBus.Gateway/Gateway.cs
```diff
@@ -50,31 +52,43 @@ protected override void Setup(FeatureConfigurationContext context)
...
-            context.Pipeline.Register("RouteToGateway", b => new RouteToGatewayBehavior(gatewayInputAddress), "Reroutes gateway messages to the gateway");
+            var retryPolicy = context.Settings.GetOrDefault<Func<IncomingMessage, Exception, int, TimeSpan>>("Gateway.Retries.RetryPolicy");
+
+            context.AddSatelliteReceiver("Gateway", gatewayInputAddress, requiredTransactionSupport, PushRuntimeSettings.Default,
+                (config, errorContext) => GatewayRecoverabilityPolicy.Invoke(errorContext, retryPolicy, config),
+                (builder, messageContext) => SendMessage(context, gatewayInputAddress, channelManager, builder, channelSenderFactory, messageContext));
 
+            context.Pipeline.Register("RouteToGateway", b => new RouteToGatewayBehavior(gatewayInputAddress), "Reroutes gateway messages to the gateway");
```
danielmarbach: You could use here the overload without the builder. Just new up the behavior

src/NServiceBus.Gateway/Gateway.cs
```diff
@@ -50,31 +52,43 @@ protected override void Setup(FeatureConfigurationContext context)
...
 
-            var retryPolicy = context.Settings.HasSetting("Gateway.Retries.RetryPolicy") ? context.Settings.Get<Func<IncomingMessage, Exception, int ,TimeSpan>>("Gateway.Retries.RetryPolicy") : DefaultRetryPolicy.BuildWithDefaults();
+            var requiredTransactionSupport = context.Settings.GetRequiredTransactionModeForReceives();
 
-            gatewayPipeline.Register(new RetryFailedGatewayMessagesBehavior.Registration(gatewayInputAddress, retryPolicy));
-            gatewayPipeline.Register("GatewaySendProcessor", b => new GatewaySendBehavior(gatewayInputAddress, channelManager, new MessageNotifier(), b.Build<IDispatchMessages>(), context.Settings, CreateForwarder(channelSenderFactory, b.BuildAll<IDataBus>()?.FirstOrDefault()), GetConfigurationBasedSiteRouter(context)), "Processes messages to be sent to the gateway");
-            context.Pipeline.Register("RouteToGateway", b => new RouteToGatewayBehavior(gatewayInputAddress), "Reroutes gateway messages to the gateway");
+            var retryPolicy = context.Settings.GetOrDefault<Func<IncomingMessage, Exception, int, TimeSpan>>("Gateway.Retries.RetryPolicy");
```
danielmarbach: Not 100 % sure but I think a get would be enough since u set a default

src/NServiceBus.Gateway/Gateway.cs
danielmarbach: The lines below could probably also use the overload which accepts a behavior instance

src/NServiceBus.Gateway/Gateway.cs
```diff
@@ -50,31 +52,43 @@ protected override void Setup(FeatureConfigurationContext context)
...
             context.Pipeline.Register("GatewayIncomingBehavior", typeof(GatewayIncomingBehavior), "Extracts gateway related information from the incoming message");
             context.Pipeline.Register("GatewayOutgoingBehavior", typeof(GatewayOutgoingBehavior), "Puts gateway related information on the headers of outgoing messages");
+
             context.RegisterStartupTask(b => new GatewayReceiverStartupTask(channelManager, channelReceiverFactory, GetEndpointRouter(context), b.Build<IDispatchMessages>(), b.Build<IDeduplicateMessages>(), b.BuildAll<IDataBus>()?.FirstOrDefault(), gatewayInputAddress));
         }
 
+        static Task SendMessage(FeatureConfigurationContext context, string gatewayInputAddress, IManageReceiveChannels channelManager, IBuilder builder, Func<string, IChannelSender> channelSenderFactory, MessageContext messageContext)
+        {
+            var sendBehavior = new GatewaySendBehavior(
+                gatewayInputAddress,
```
danielmarbach: Do we need to new up that object every time we send?

danielmarbach:
@marcinhoppe I made a few comments. I might find time address them

MarcinHoppe:
@danielmarbach Thanks, I'll look at these and ping you when implemented. Smoke tests went fine, all seemed to work.

src/NServiceBus.Gateway.Tests/APIApprovals.Approve.approved.txt
MarcinHoppe: To be consistent, we would also need to rename the `Retries` method to `DelayedRetries`. I think these names would be more intention revealing given recent changes in the core recoverability APIs.

@danielmarbach Do you agree to rename both methods?

MarcinHoppe:
@danielmarbach @andreasohlund Improved stuff a little bit. Would you mind taking a look?

andreasohlund:
LGTM

MarcinHoppe:
Only outstanding item is if we should align the retry configuration API for the gateway with the core more than it is now?

src/NServiceBus.Gateway/Gateway.cs
```diff
@@ -50,28 +51,33 @@ protected override void Setup(FeatureConfigurationContext context)
...
+            var sender = new GatewayMessageSender(
+                gatewayInputAddress,
+                channelManager,
+                new MessageNotifier(),
+                context.Settings.LocalAddress(),
+                GetConfigurationBasedSiteRouter(context));
+
+            context.AddSatelliteReceiver("Gateway", gatewayInputAddress, requiredTransactionSupport, PushRuntimeSettings.Default,
+                (config, errorContext) => GatewayRecoverabilityPolicy.Invoke(errorContext, retryPolicy, config),
+                (builder, messageContext) => sender.SendToDestination(messageContext, builder.Build<IDispatchMessages>(), CreateForwarder(channelSenderFactory, builder.BuildAll<IDataBus>()?.FirstOrDefault())));
```
danielmarbach: The dispatcher should be stateless by design. No need to create a new one for each send. You could pass it into the ctor of the GatewayMessageSender. What about the forwarder, I have a hunch that can be a singleton too?

andreasohlund: >The dispatcher should be stateless by design

But we need to get it from the builder so there is no other way to do it? (or did you mean the "forwarder")

MarcinHoppe: The forwarder needs `IDataBus` which needs to be resolved using the builder. I had an alternative version that registered `GatewayMessageSender` singleton with the container and then resolve in `onMessage` but I think the current approach is more consistent with other satellites in the core.

danielmarbach:
Pulling in. We can fight over https://github.com/Particular/NServiceBus.Gateway/pull/54#issuecomment-238485985 in a dedicated PR ;)

Issue #55, opened 2016-08-01
NSB.Gateway 1.0 will not accept messages from NSB 3.x


DavidBoike:
A V3 endpoint cannot send messages to a Gateway running NSB 5.x and NSB.Gateway 1.0.1.

The V3 client gets `400 Bad Request`.

The V5 server shows this:

> WARN  NServiceBus.Gateway.Channels.Http.HttpChannelReceiver Cannot process HTTP request from [fe80::b061:f586:e59e:6bab%4]:57604. Reason: Invalid CallType 'Submit'. CallTypes supported 'SingleCallSubmit, SingleCallDatabusProperty'.

In V3, the Gateway uses these call types:

    public enum CallType
    {
        Submit,
        Ack,
        DatabusProperty
    }

But in CoreV5/NSB.Gateway 1.0, it seems to have changed drastically:

    public enum CallType
    {
        /// <summary>
        /// Default request type.
        /// </summary>
        SingleCallSubmit,
        /// <summary>
        /// Request type for Databus properties.
        /// </summary>
        SingleCallDatabusProperty
    }

It seems "single call" means that Acks are not required. In V3, it does a `Send`, then sends any data bus properties, and then ends with an `Ack`.

Calling on @johnsimons as you touched Gateway code in V3, otherwise @SimonCropp and @andreasohlund may be the only ones with enough historical perspective on this. (Outside Udi of course.)

Related to [Desk Case](https://nservicebus.desk.com/agent/case/18676).

DavidBoike:
Created a WIP PR in #56 that may be a potential fix, although I don't fully understand the implications.

johnsimons:
@DavidBoike I am back next Monday, can this wait?

johnsimons:
@DavidBoike just a few observations, does it work from v3 -> v4 -> v5 ?

DavidBoike:
@johnsimons Yes I think it can wait. In the meantime, the customer is trying out a NuGet package created from the PR in #56 to see if that works for them.

DavidBoike:
Pinged the customer to find out if the prerelease package worked.

@johnsimons Could you review the PR in #56 and see what you think? It seems like the Ack was designed to say any DataBus stuff was done. Would it perhaps work just fine, as long as you also don't use the DataBus with it?

johnsimons:
@DavidBoike the issue is that a customer needs to update from v3->v4 and the go from v4->v5.

[...]

PR #56, opened 2016-08-01
[WIP] Allow response to V3 gateway requests

DavidBoike:
**DO NOT MERGE - EXPLORATORY ONLY**

Connects to #55

The additional enumeration members allow the gateway to treat a V3 `Submit` as a current version `SingleCallSubmit`, and ignore an `Ack` completely.

I don't know if this will have any other undesirable effects, as I don't fully understand how the change from V3's gateway model to the current evolved.

### DIFF
--- a/src/NServiceBus.Gateway/CallType.cs
+++ b/src/NServiceBus.Gateway/CallType.cs
@@ -1,5 +1,7 @@
 namespace NServiceBus.Gateway
 {
+    using System;
+
     /// <summary>
     /// received request type.
     /// </summary>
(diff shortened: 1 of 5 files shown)


src/NServiceBus.Gateway/CallType.cs
```diff
@@ -12,6 +14,17 @@ public enum CallType
         /// <summary>
         /// Request type for Databus properties.
         /// </summary>
-        SingleCallDatabusProperty
+        SingleCallDatabusProperty,
```
johnsimons: Shouldn't this one also map to the v3 equivalent (DatabusProperty) ?

DavidBoike:
Closing as Won't Fix. I don't like the idea of reintroducing API members that were specifically obsoleted and then removed in previous versions.

Also, V3 is technically unsupported, and customer decided to have client sites upgrade to recent version anyway.

Issue #12, opened 2014-10-18
Add support for reply address per gateway channel


SimonCropp:
Current gateway configuration allows only for a single reply to address. When using the built-in Http channel, this address is also used for hosting the HttpListener.

When the gateway is hosted behind a load balancer, the public address of the gateway can be different from the local address on the machine hosting the actual gateway endpoint.

Having separate reply address config for each channel will decouple the HttpListener from the reply address and allow more flexibility when additional gateway channels (like SQS, ActiveMQ, Azure) are available later on.

http://tech.groups.yahoo.com/group/nservicebus/message/20000

johnsimons:
hi @chrisbednarski,
Finally going through some old request for features on the gateway code and I come across this one raised by you.

Are you proposing to add an extra configuration settings to the `ChannelConfig` to set the return address?
Would we return this address in a http header ? Which one? 

is this it?

chrisbednarski:
Hi @johnsimons 
Yes, this is all about an extra configuration settings on the `Channels` config section.

We currently achieve our goals by adding a dummy channel and tagging it as the default.

```
<GatewayConfig>
	<Sites>
  	</Sites>
	<Channels>
		<Channel Address="https://some.address.com/" ChannelType="Http" Default="true" />
		<Channel Address="http://+:80/" ChannelType="Http" />
	</Channels>
</GatewayConfig>
```

Load balancer passes the HTTPS traffic onto the HTTP port 80.
`Bus.Reply` uses the address from the default channel and everything falls into place.

It would be better if I didn't have to do this workaround.

chrisbednarski:
I think it's the `Headers.OriginatingSite` and can be found [here](https://github.com/Particular/NServiceBus.Gateway/blob/develop/src/NServiceBus.Gateway/Sending/GatewaySender.cs#L93)

chrisbednarski:
there's an old pull request https://github.com/Particular/NServiceBus/pull/1401

johnsimons:
Thank you @chrisbednarski 

Looking at the workaround you came up with, I actually do not think is a bad one at all, which makes me wonder if we should just document it as the way to go about when behind a load balancer, thoughts? 

I looked at the https://github.com/Particular/NServiceBus/pull/1401 and there is a bit of changes just to make it work, I understand why, but I was wondering if we could instead add the `ReplyAddress` to the `SiteConfig`, I know it may feel wrong, and maybe a maintenance nightmare but just wondering, thoughts?

[...]

PR #66, opened 2017-01-17
Support for listening to urls with wildcards

janovesk:
### Problem description

When NSB.Gateway is used in an environment where there's no control over the channel address is possible, at the time when address is registered with URLACL, it must be registered with a wildcard.

### Use case

One of the customers was trying to host the `RemoteSite` equivalent of the [NSB.Gateway sample](https://docs.particular.net/samples/gateway/) in a Service Fabric (SF) sample. Problem with that approach is the URL registered for a channel cannot be in a URI format as below:

```xml
    <Channel Address="http://localhost:25899/RemoteSite/"
             ChannelType="Http"
             Default="true"/>
```

since requests forwarded to nodes could use FQDN or an IP address. Therefore, the proper registration should be 

```xml
    <Channel Address="http://+:25899/RemoteSite/"
             ChannelType="Http"
             Default="true"/>
```

URL with a wildcard fails, due to NSB.Gateway internal usage of `Uri` to parse the schema. This causes NSB.Gateway to fail when a wildcard URL is registered.

### Solution

This PR replaces `Uri` usage with a primitive `string` considering that `Shema` part is still used as a plain string, the `Uri` is not even needed.

This has been confirmed with a customer to work (using PR nuget package).

@Particular/nservicebus-maintainers please review.

### DIFF
--- a/src/NServiceBus.Gateway/GatewayHttpListenerInstaller.cs
+++ b/src/NServiceBus.Gateway/GatewayHttpListenerInstaller.cs
@@ -48,8 +48,8 @@ netsh http add urlacl url={{http://URL:PORT/[PATH/] | https://URL:PORT/[PATH/]}}
             {
                 if (receiveChannel.Type.ToLower() != "http") continue;
 
-                var uri = new Uri(receiveChannel.Address);
-                if (!uri.Scheme.StartsWith("http", StringComparison.InvariantCultureIgnoreCase))
+                var uri = receiveChannel.Address;
+                if(!uri.StartsWith("http", StringComparison.InvariantCultureIgnoreCase))
                 {
                     continue;
                 }
@@ -69,7 +69,7 @@ netsh http add urlacl url={1} user=""{0}""", uri, identity);
             return Task.FromResult(0);
         }
 
-        static void StartNetshProcess(string identity, Uri uri)
+        static void StartNetshProcess(string identity, string uri)
         {
             var startInfo = new ProcessStartInfo
             {


SeanFeldman:
Confirmed with the customer that wildcard allows Gateway to be used with SF. :shipit:

SeanFeldman:
@janovesk anything else is needed for this to go in?

andreasohlund:
@janovesk do you need any help with doco and releasing this?

SeanFeldman:
Can help if needed. Would like the customer to use an official build, and not a partisan provided one 😃

janovesk:
Hmmm, I have a concern. This would break `.Reply()` unless the user uses [this workaround](https://github.com/Particular/NServiceBus.Gateway/issues/12#issuecomment-73604791). 

```
<GatewayConfig>
    <Sites>
    </Sites>
    <Channels>
        <Channel Address="https://some.address.com/" ChannelType="Http" Default="true" />
        <Channel Address="http://+:80/" ChannelType="Http" />
    </Channels>
</GatewayConfig>
```

Looks like a reasonable fine thing to do, we can just document that if you have a channel with a wildcard, you must also provide a non-wildcard default to make sure `.Reply()` will work.

Thoughts, @SeanFeldman @andreasohlund @bording ?

andreasohlund:
Can we check and require it on startup? (and perhaps ask for a config.IWontDoAnyReplies() if the users don't want to add it)

janovesk:
Yeah, that should be fairly trivial to add. I'll give it a go.

SeanFeldman:
@janovesk when we have that, I'd like to provide a temp build to the customer to try out in his solution to confirm behavior in a cloud deployed SF cluster.

PR #72, opened 2017-08-24
Update to Core v7 alpha

andreasohlund:


andreasohlund:
@Particular/nservicebus-maintainers should be good to go

PR #73, opened 2017-08-28
Add new code first api

andreasohlund:
This PR updates to latestes core v7. Since `IProvideConfiguration` and `IConfigurationSource` was removed from the core I had to add a new code first api to not requires us to force users back to app.config only.

Doco PR will come once we're happy with the api.


src/NServiceBus.Gateway.AcceptanceTests/EndpointTemplates/DefaultServerWithNoStorage.cs
```diff
@@ -0,0 +1,42 @@
+﻿namespace NServiceBus.AcceptanceTests.EndpointTemplates
```
SimonCropp: redundant namespaces

src/NServiceBus.Gateway.AcceptanceTests/NServiceBusAcceptanceTest.cs
```diff
@@ -13,6 +13,10 @@ public abstract class NServiceBusAcceptanceTest
         [SetUp]
         public void SetUp()
         {
+            // Hack: prevents SerializationException ... Type 'x' in assembly 'y' is not marked as serializable.
+            // https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/mitigation-deserialization-of-objects-across-app-domains
+            System.Configuration.ConfigurationManager.GetSection("X");
```
SimonCropp: didnt we fix this in core?

src/NServiceBus.Gateway/Settings/GatewaySettings.cs
```diff
@@ -76,9 +87,151 @@ public void DisableRetries()
...
+                Key = siteKey,
+                LegacyMode = legacyMode
+            };
+
+            if (config.GetSettings().TryGet(out List<Site> sites))
+            {
+                sites.Add(site);
+            }
+
+            config.GetSettings().Set<List<Site>>(new List<Site>
```
SimonCropp: assign a local for ` config.GetSettings()`

src/NServiceBus.Gateway/Settings/GatewaySettings.cs
```diff
@@ -76,9 +87,151 @@ public void DisableRetries()
...
+                Type = type,
+                Default = isDefault
+            };
+
+            if (config.GetSettings().TryGet(out List<ReceiveChannel> channels))
+            {
+                channels.Add(channel);
+            }
+
+            config.GetSettings().Set<List<ReceiveChannel>>(new List<ReceiveChannel>
```
SimonCropp: local for settings

src/NServiceBus.Gateway/Utils/Guard.cs
```diff
@@ -0,0 +1,93 @@
...
+    using System.Linq;
+    using System.Reflection;
+
+    static class Guard
+    {
+// ReSharper disable UnusedParameter.Global
+        public static void TypeHasDefaultConstructor(Type type, string argumentName)
+        {
+            if (type.GetConstructors(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic)
+                .All(ctor => ctor.GetParameters().Length != 0))
```
SimonCropp: `ctor.GetParameters().Length != 0)` doesnt this assert that all constructors have zero params? not that one does?

src/NServiceBus.Gateway.AcceptanceTests/NServiceBusAcceptanceTest.cs
andreasohlund: It was fixed in the shipped acceptance tests but the GW only runs it own test so I needed to copy it over

https://github.com/Particular/NServiceBus/blob/ed11623e2d7f6fdedee1b756c800f06606a88bfb/src/NServiceBus.AcceptanceTests/NServiceBusAcceptanceTest.cs#L21

src/NServiceBus.Gateway/Utils/Guard.cs
andreasohlund: Removed the method since we don't use it but since it was copied from the core there might be a problem there?

andreasohlund:
@SimonCropp addressed your review comments

SimonCropp:
@andreasohlund if u merge and promote i will handle the docs

SimonCropp:
@andreasohlund actually it would be better for you to write the upgrade guide

andreasohlund:
Yea, I'll deal with it!

PR #74, opened 2017-08-29
Hotfix 2.0.1

janovesk:
Support for listening to wildcard http uris

### DIFF
--- a/src/NServiceBus.Gateway.AcceptanceTests/When_listening_on_wildcard_uri.cs
+++ b/src/NServiceBus.Gateway.AcceptanceTests/When_listening_on_wildcard_uri.cs
@@ -1,0 +1,46 @@
+﻿namespace NServiceBus.AcceptanceTests.Gateway
+{
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using Config;
+    using EndpointTemplates;
+    using Features;
+    using NUnit.Framework;
+
+    public class When_listening_on_wildcard_uri : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public void should_throw_exception_if_wildcard_channel_is_used_for_replies()
+        {
+            Assert.That(async () =>
+            {
+                await Scenario.Define<ScenarioContext>()
+                    .WithEndpoint<EndpointWithWildCardUriAsDefault>(e => e.When(b => Task.FromResult(0)))
+                    .Run();
+            }, Throws.Exception.InnerException.InnerException.With.Message.Contains("An extra channel with a fully qualified non-wildcard url must be configured as default for Reply routing."));
+        }
+
+        class EndpointWithWildCardUriAsDefault : EndpointConfigurationBuilder
+        {
+            public EndpointWithWildCardUriAsDefault()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    c.EnableFeature<Gateway>();
+                })
+                    .WithConfig<GatewayConfig>(c =>
+                    {
+                        c.Channels = new ChannelCollection
+                        {
+                            new ChannelConfig
+                            {
+                                Address = "http://*:25699/",
+                                ChannelType = "http",
+                                Default = true
+                            }
+                        };
+                    });
+            }
+        }
+    }
+}
--- a/src/NServiceBus.Gateway.AcceptanceTests/When_listening_on_wildcard_uri.cs
+++ b/src/NServiceBus.Gateway.AcceptanceTests/When_listening_on_wildcard_uri.cs
@@ -34,7 +34,7 @@
                         {
                             new ChannelConfig
                             {
-                                Address = "http://*:25699/",
+                                Address = "http://+:25699/",
                                 ChannelType = "http",
                                 Default = true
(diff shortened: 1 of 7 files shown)


src/NServiceBus.Gateway/Gateway.cs
```diff
@@ -94,10 +94,23 @@ static void RegisterChannels(FeatureConfigurationContext context, IManageReceive
...
+                var repliesDisabled = context.Settings.GetOrDefault<bool>("Gateway.DisableReplyChannelCheck");
+                if (repliesDisabled == false)
+                {
+                    CheckForNonWildcardDefaultChannel(channelManager);
+                }
+                RegisterHttpListenerInstaller(context, channelManager);
             }
+        }
 
-            var enableHttpListener = !usingCustomChannelProviders;
```
SzymonPobiega: I think it would be better to set the flag for the installer in some object and resolve that object from the container, like it is done in NH. Various containers might have different behavior when re-registering thins in the container and the installers are first registered by convention in the core.

janovesk:
[Doco](https://github.com/Particular/docs.particular.net/pull/3159) ready

src/NServiceBus.Gateway.AcceptanceTests/When_listening_on_wildcard_uri.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        [Test]
+        public void should_throw_exception_if_wildcard_channel_is_used_for_replies()
+        {
+            Assert.That(async () =>
+            {
+                await Scenario.Define<ScenarioContext>()
+                    .WithEndpoint<EndpointWithWildCardUriAsDefault>(e => e.When(b => Task.FromResult(0)))
+                    .Run();
+            }, Throws.Exception.InnerException.InnerException.With.Message.Contains("An extra channel with a fully qualified non-wildcard url must be configured as default for Reply routing."));
```
andreasohlund: Just noting that when merging to develop `InnerException.InnerException` needs to be removed due to changes in the att framework

src/NServiceBus.Gateway/Gateway.cs
```diff
@@ -89,15 +93,26 @@ static void RegisterChannels(FeatureConfigurationContext context, IManageReceive
...
+        }
+
+        static void CheckForNonWildcardDefaultChannel(IManageReceiveChannels channelManager)
+        {
+            var defaultChannel = channelManager.GetDefaultChannel();
+            if (defaultChannel.Address.Contains("*") || defaultChannel.Address.Contains("+"))
             {
-                channelReceiverFactory = s => new ChannelReceiverFactory(typeof(HttpChannelReceiver)).GetReceiver(s);
-                channelSenderFactory = s => new ChannelSenderFactory(typeof(HttpChannelSender)).GetSender(s);
+                throw new Exception($"Listening on {defaultChannel.Address} with a wildcard uri. An extra channel with a fully qualified non-wildcard url must be configured as default for Reply routing.");
```
andreasohlund: How about:

>Default channel {defaultChannel.Address} is using a wildcard uri. Please add an extra channel with a fully qualified non-wildcard url in order for replies to be transmitted properly

src/NServiceBus.Gateway/Gateway.cs
```diff
@@ -89,15 +93,26 @@ static void RegisterChannels(FeatureConfigurationContext context, IManageReceive
...
+
+            var installerSettings = context.Settings.Get<InstallerSettings>();
+            installerSettings.ChannelManager = channelManager;
+            installerSettings.Enabled = true;
+        }
+
+        static void CheckForNonWildcardDefaultChannel(IManageReceiveChannels channelManager)
+        {
+            var defaultChannel = channelManager.GetDefaultChannel();
+            if (defaultChannel.Address.Contains("*") || defaultChannel.Address.Contains("+"))
```
andreasohlund: Don't we have to check if this is a http channel?

janovesk: Nah, it is implicit. This method is only called if you are running with the default http channel manager.

src/NServiceBus.Gateway/Gateway.cs
```diff
@@ -89,15 +93,26 @@ static void RegisterChannels(FeatureConfigurationContext context, IManageReceive
...
             {
-                channelReceiverFactory = s => new ChannelReceiverFactory(typeof(HttpChannelReceiver)).GetReceiver(s);
-                channelSenderFactory = s => new ChannelSenderFactory(typeof(HttpChannelSender)).GetSender(s);
+                throw new Exception($"Default channel {defaultChannel.Address} is using a wildcard uri. Please add an extra channel with a fully qualified non-wildcard url in order for replies to be transmitted properly.");
```
andreasohlund: Nitpick: We're mixing Uri and Url in the message

bording:
@janovesk Looking the state of the develop branch, while #66 is in develop, it appears that there were more changes in the 2.0.1 release that don't appear to have made it in.

Does `hotfix-2.0.1` need to be merged to develop as well?

PR #77, opened 2017-09-04
Csharp71

andreasohlund:


andreasohlund:
Trivial change and we need to get the release branch back to green so self merging this

PR #80, opened 2017-10-09
prevent fody from propagating

SimonCropp:
so the produced nuget will not have a dependency on fody


andreasohlund:
@SimonCropp I retargeted against the `release-3.0` branch

PR #86, opened 2017-12-08
[WIP] Allow configuration of reverse proxy friendly headers

### LINKED ISSUE #84: Gateway header value passing is incompatible with Cloudflare
Cloudflare does not support dots `.` in header keys which is used on NServiceBus headers. To make it compatible with Cloudflare either the header passing must be changed or should be customizable.

welshdave:
Added the ability to configure remote sites as being behind reverse proxies (such as nginx and cloudflare) that don't allow headers with periods in them. I appreciate there might not be too many people who ever need to do this so I've tried to make it as unobtrusive as possible. fixes #84

### DIFF
--- a/src/NServiceBus.Gateway.Tests/APIApprovals.Approve.approved.txt
+++ b/src/NServiceBus.Gateway.Tests/APIApprovals.Approve.approved.txt
@@ -62,6 +62,8 @@ namespace NServiceBus.Config
         public string Key { get; set; }
         [System.Configuration.ConfigurationPropertyAttribute("LegacyMode", DefaultValue=true, IsKey=false, IsRequired=false)]
         public bool LegacyMode { get; set; }
+        [System.Configuration.ConfigurationPropertyAttribute("UsesReverseProxy", DefaultValue=false, IsKey=false, IsRequired=false)]
+        public bool UsesReverseProxy { get; set; }
     }
 }
 namespace NServiceBus.Features
@@ -130,6 +132,7 @@ namespace NServiceBus.Gateway.Routing
         public NServiceBus.Gateway.Channels.Channel Channel { get; set; }
         public string Key { get; set; }
         public bool LegacyMode { get; set; }
+        public bool UsesReverseProxy { get; set; }
     }
 }
 namespace NServiceBus
@@ -141,7 +144,7 @@ namespace NServiceBus
     public class GatewaySettings
     {
         public void AddReceiveChannel(string address, string type = "http", int maxConcurrency = 1, bool isDefault = False) { }
-        public void AddSite(string siteKey, string address, string type = "http", bool legacyMode = False) { }
+        public void AddSite(string siteKey, string address, string type = "http", bool legacyMode = False, bool usesReverseProxy = False) { }
         public void ChannelFactories(System.Func<string, NServiceBus.Gateway.IChannelSender> senderFactory, System.Func<string, NServiceBus.Gateway.IChannelReceiver> receiverFactory) { }
         public void CustomRetryPolicy(System.Func<NServiceBus.Transport.IncomingMessage, System.Exception, int, System.TimeSpan> customRetryPolicy) { }
         public void DisableRetries() { }
--- a/src/NServiceBus.Gateway.Tests/APIApprovals.Approve.approved.txt
+++ b/src/NServiceBus.Gateway.Tests/APIApprovals.Approve.approved.txt
@@ -62,8 +62,6 @@ namespace NServiceBus.Config
         public string Key { get; set; }
         [System.Configuration.ConfigurationPropertyAttribute("LegacyMode", DefaultValue=true, IsKey=false, IsRequired=false)]
         public bool LegacyMode { get; set; }
-        [System.Configuration.ConfigurationPropertyAttribute("UsesReverseProxy", DefaultValue=false, IsKey=false, IsRequired=false)]
-        public bool UsesReverseProxy { get; set; }
     }
 }
 namespace NServiceBus.Features
--- a/src/NServiceBus.Gateway/Config/SiteConfig.cs
+++ b/src/NServiceBus.Gateway/Config/SiteConfig.cs
@@ -72,22 +72,6 @@ namespace NServiceBus.Config
                 this["LegacyMode"] = value;
             }
         }
-
-        /// <summary>
-        /// Is the site behind a reverse proxy 
-        /// </summary>
-        [ConfigurationProperty("UsesReverseProxy", IsRequired = false, DefaultValue = false, IsKey = false)]
-        public bool UsesReverseProxy
-        {
-            get
-            {
-                return (bool) this["UsesReverseProxy"]; 
-            }
-            set
-            {
-                this["UsesReverseProxy"] = value;
-            }
-        }
     }
 }
 #endif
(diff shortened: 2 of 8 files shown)


src/NServiceBus.Gateway/Config/SiteConfig.cs
```diff
@@ -72,6 +72,22 @@ public bool LegacyMode
                 this["LegacyMode"] = value;
             }
         }
+
+        /// <summary>
+        /// Is the site behind a reverse proxy 
+        /// </summary>
+        [ConfigurationProperty("UsesReverseProxy", IsRequired = false, DefaultValue = false, IsKey = false)]
+        public bool UsesReverseProxy
```
andreasohlund: As we're moving towards a config model with code api's only I would prefer to not expose this here? (and just go with the code first api)

andreasohlund:
@welshdave what version of the gateway are you using? (since the PR is targeting develop this will likely go into a 3.1.0 targeting v7, is that ok with you?)

src/NServiceBus.Gateway/Config/SiteConfig.cs
welshdave: Yeah, I'm happy to remove that, we're using code api's to configure anyway. Will give it a bit to see if any more feedback then update.

welshdave:
@andreasohlund We're currently using v6, but planning to update to v7 (and this getting into a version targeting v7 would make that higher priority).

src/NServiceBus.Gateway.Tests/APIApprovals.Approve.approved.txt
```diff
@@ -141,7 +142,7 @@ namespace NServiceBus
     public class GatewaySettings
     {
         public void AddReceiveChannel(string address, string type = "http", int maxConcurrency = 1, bool isDefault = False) { }
-        public void AddSite(string siteKey, string address, string type = "http", bool legacyMode = False) { }
+        public void AddSite(string siteKey, string address, string type = "http", bool legacyMode = False, bool usesReverseProxy = False) { }
```
dbelcham: While this has a default value for the new parameter in the signature, does it constitute a breaking change in the semver world? Part of me says no (users don't need to change anything because of the default value), part of me says yes because the public API is actually changing. Right now I'm leaning towards calling this a breaking change. 

@Particular/nservicebus-maintainers  your thoughts?

src/NServiceBus.Gateway/Settings/GatewaySettings.cs
```diff
@@ -83,12 +83,13 @@ public void DisableRetries()
...
         /// <param name="legacyMode">Pass `true` to set the forwarding mode for this site to legacy mode.</param>
-        public void AddSite(string siteKey, string address, string type = "http", bool legacyMode = false)
+        /// <param name="usesReverseProxy">Pass `true` if site is behind a reverse proxy that doesn't like headers with periods in them.</param>
+        public void AddSite(string siteKey, string address, string type = "http", bool legacyMode = false, bool usesReverseProxy = false)
         {
             Guard.AgainstNullAndEmpty(nameof(siteKey), siteKey);
             Guard.AgainstNullAndEmpty(nameof(address), address);
             Guard.AgainstNullAndEmpty(nameof(type), type);
-
+
```
dbelcham: 🔥 the whitespace

src/NServiceBus.Gateway.Tests/APIApprovals.Approve.approved.txt
bording: Optional parameters are baked into the call site at compile time, so this is definitely a binary breaking change.

To do this is a non-breaking way, it would need to be a new overload that adds the new parameter instead.

src/NServiceBus.Gateway/GatewayExtensions.cs
```diff
@@ -29,5 +29,19 @@ public static bool IsLegacyGatewayMessage(this IDictionary<string, string> heade
...
+
+        /// <summary>
+        ///     encode headers for reverse proxy
+        /// </summary>
+        public static IDictionary<string, string> EncodeHeadersForReverseProxy(this IDictionary<string, string> headers)
+        {
+            var encodedHeaders = new Dictionary<string, string>();
+            foreach (var header in headers)
+            {
+                encodedHeaders.Add(header.Key.Replace(".","-"),header.Value);
```
ramonsmits: The encoder is different in behavior then the decoder.

ramonsmits: Only headers starting with `NServiceBus.` should then be addressed. Also, this would mean that we *must* make sure that we always use dots in our header keys.

bording: @ramonsmits It's not that we have to use `.`, it's that we can't use `-` in anywhere in the key, or it will incorrectly be turned into `.` on the decoding side.

Given that we don't have any sort of enforcement around this, I'm not sure how tenable the approach being taken here is.

welshdave: I think it should only be updating headers that start `NServiceBus.`. I did a bit more playing around to get different headers, and it looks like my current approach won't work when using a databus; that adds headers with `-` in them and also `_`, which is another problem character for nginx.

I think that something like this could work:

If the header begins `NServiceBus.`, then replace with `NServiceBus-{base64 encoded header name}`. As long as all NServiceBus headers start with `NServiceBus.` then that should be ok. Any thoughts on that approach?

andreasohlund:
Another option would be to add an opt-in config option to send all the headers in the payload instead? (since we don't really need them to be http headers unless the GW is used to provide integration over http)

welshdave:
@andreasohlund That could work. A first pass at base64 encoding is failing (again when using a databus), so I'm going to dig into that a bit, but will look at this option as well.

dbelcham:
@welshdave we talked through this quite thoroughly today and came to the conclusion that there is a larger fundamental architectural issue in the Gateway. Realistically, all of the NServiceBus messages don't belong as HTTP headers. Instead they should be part of the HTTP body.

The challenge that this presents is that simply moving all of the headers to the body would break our "wire compatibility" (the idea that messages sent are backwards compatible with older versions of NServiceBus. So to solve this we need to have a solution where the future versions of the Gateway will send messages with headers in the body, and receive and correctly process messages that have headers either in the body or as HTTP headers. We also think that putting messages into the body should be an option that you have to opt-in on since due to backwards compatibility of code (expected behaviors from the Gateway) and due to existing use-cases where HTTP headers will be used.

So...with all of that said, is this something you're interested in doing still? It is very different from the PR you have raised here which is why we wanted to bring it up with you. If you want to have a detailed chat with someone about the ideas, @andreasohlund has offered to jump on a call with you at your convenience. Just let us know and we'll make that happen.

welshdave:
Yeah, I'm happy with doing that. What's the best way to set up a call with @andreasohlund?

andreasohlund:
Email me at andreas dot ohlund at particular net and we'll work something
out

andreasohlund:
@welshdave based on our discussions should we close this? (can you link to your new gateway channel)

welshdave:
Addressing this requirement with a new channel implementation: https://github.com/welshdave/NServiceBus.Gateway.Channels.HttpVNext

PR #87, opened 2017-12-14
Add gateway storage acceptance test package

janovesk:
To allow downstream persisters to test the gateway storage

* To enable downstreams to setup things like tables etc the acceptance tests had to enable the installers. This highlighted a bug where the GW code assumed the feature to always be enabled. This is no longer true and since[ the core will only set defaults for enabled features](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Features/FeatureActivator.cs#L54) this turned out to be a bug since installers always run no matter what. This bug is fixed in this PR. The ex was `System.Collections.Generic.KeyNotFoundException : The given key (NServiceBus.Gateway.Installer.InstallerSettings) was not present in the dictionary.`

### DIFF
--- a/src/NServiceBus.Gateway.AcceptanceTests/NServiceBus.Gateway.AcceptanceTests.csproj
+++ b/src/NServiceBus.Gateway.AcceptanceTests/NServiceBus.Gateway.AcceptanceTests.csproj
@@ -41,4 +41,9 @@
     <_PackageFiles Remove="**\InMemoryTestSuiteConstaints.cs" />
   </ItemGroup>
 
+  <ItemGroup>
+    <_PackageFiles Remove="GatewayEndpoint.cs" />
+    <_PackageFiles Remove="GatewayEndpointWithNoStorage.cs" />
+  </ItemGroup>
+
 </Project>
--- a/src/NServiceBus.Gateway.AcceptanceTests/NServiceBus.Gateway.AcceptanceTests.csproj
+++ b/src/NServiceBus.Gateway.AcceptanceTests/NServiceBus.Gateway.AcceptanceTests.csproj
@@ -41,9 +41,4 @@
     <_PackageFiles Remove="**\InMemoryTestSuiteConstaints.cs" />
   </ItemGroup>
 
-  <ItemGroup>
-    <_PackageFiles Remove="GatewayEndpoint.cs" />
-    <_PackageFiles Remove="GatewayEndpointWithNoStorage.cs" />
-  </ItemGroup>
-
 </Project>
--- a/src/NServiceBus.Gateway.AcceptanceTests/NServiceBus.Gateway.AcceptanceTests.csproj
+++ b/src/NServiceBus.Gateway.AcceptanceTests/NServiceBus.Gateway.AcceptanceTests.csproj
@@ -42,12 +42,4 @@
     <_PackageFiles Remove="**\InMemoryTestSuiteConstaints.cs" />
   </ItemGroup>
 
-  <ItemGroup>
-    <_PackageFiles Remove="GatewayTestSuiteConstraints.cs" />
-    <_PackageFiles Remove="IConfigureGatewayPersitenceExecution.cs" />
-    <_PackageFiles Remove="IGatewayTestSuiteConstraints.cs" />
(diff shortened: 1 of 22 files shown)


janovesk:
Connects to https://github.com/Particular/DotNetCoreLaunch/issues/132

src/NServiceBus.Gateway.AcceptanceTests/NServiceBus.Gateway.AcceptanceTests.csproj
```diff
@@ -19,6 +19,23 @@
...
+    <PackageId>NServiceBus.Gateway.AcceptanceTests.Sources</PackageId>
+    <Description>Acceptance tests for NServiceBus gateway functionality</Description>
+    <IsPackable>true</IsPackable>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <_PackageFiles Include="**\*.cs">
+      <BuildAction>Compile</BuildAction>
+      <PackagePath>content\App_Packages\NSB.GW.AcceptanceTests\;contentFiles\cs\any\NSB.GW.AcceptanceTests\</PackagePath>
```
bording: Would `Gateway` be better than `GW` here?

src/NServiceBus.Gateway.AcceptanceTests/NServiceBus.Gateway.AcceptanceTests.csproj
```diff
@@ -19,6 +19,23 @@
...
+    </_PackageFiles>
+    <_PackageFiles Remove="**\obj\**\*.cs" />    
+    <_PackageFiles Remove="AssemblyInfo.cs" />
```
bording: There is no `AssemblyInfo.cs` file in this is project, so this line isn't needed.

src/NServiceBus.Gateway.AcceptanceTests/NServiceBus.Gateway.AcceptanceTests.csproj
```diff
@@ -19,6 +19,23 @@
     <PackageReference Include="NServiceBus.Callbacks" Version="3.0.0-beta0003" />
     <PackageReference Include="NUnit" Version="3.7.1" />
     <PackageReference Include="NUnit3TestAdapter" Version="3.8.0-alpha1" />
+    <PackageReference Include="Particular.Packaging" Version="0.1.0" PrivateAssets="All" />
```
bording: The rest of the package references either need `PrivateAssets="All"`, or a proper version range for the dependencies that should be on the package.

Private:
- Microsoft.NET.Test.Sdk
- NUnit
- NUnit3TestAdapter

Range:
- NServiceBus.AcceptanceTesting:  `[7.0.0-beta0012, 8.0.0)`
- NServiceBus.Callbacks: `[3.0.0-beta0003, 4.0.0)`

bording: NServiceBus.AcceptanceTesting brings in a transitive NUnit dependency, so this project could drop the NUnit dependency.

Alternatively, if this project bumped up the version of NUnit it uses, you could argue that the package would then need a public NUnit as well, to ensure the proper minimum version is picked up.

andreasohlund:
@janovesk this PR needs to be retargeted against `release-3.0` since we need to get this into the downstreams before they can ship their RC's

andreasohlund:
@bording addressed your comments

andreasohlund:
@timbussmann can you see if I got the "extensibility" stuff right? (the partials etc)

andreasohlund:
Smoke tested the latest PR build in NHibernate and it runs green locally https://github.com/Particular/NServiceBus.NHibernate/pull/317

bording:
With the changes I pushed up, from a packaging standpoint this looks 👍 

@timbussmann should definitely look over the test configuration extensibility stuff before merging, though!

andreasohlund:
The tests now runs fine on RavenDB as well https://github.com/Particular/NServiceBus.RavenDB/pull/382

PR #90, opened 2018-01-26
Prepare RC1

timbussmann:


bording:
This made a change that is wrong. `NServiceBus.Gateway.AcceptanceTests` projects needs the version ranges because we ship a source package from there.

PR #104, opened 2018-10-03
Throw when sending to non configured site

### LINKED ISSUE #102: Sending to unknown sites silently fails
If users forget to call `.AddSite("B")` calls to `.SendToSites("A","B")` will silently just ignore it. 

We should throw on send when this is true.

andreasohlund:
Fixes #102


andreasohlund:
Tests are failing because some agents have bad urlacls on them. Working to fix that, should be ready for review still

src/NServiceBus.Gateway/Routing/RouteToGatewayBehaviour.cs
```diff
@@ -1,23 +1,44 @@
...
         public override Task Invoke(IRoutingContext context, Func<Task> next)
         {
-
             if (context.Extensions.TryGet(out RouteThroughGateway _))
             {
+                if (context.Message.Headers.TryGetValue(Headers.DestinationSites, out var siteKeyHeader))
+                {
+                    var siteKeys = siteKeyHeader.Split(',');
+
+                    var unknownSites = new List<string>();
```
bording: Can we change this so we're only allocating a collection if we actually have an unknown site? Otherwise we're always creating this and then just throwing it away for each message.

src/NServiceBus.Gateway/Routing/RouteToGatewayBehaviour.cs
```diff
@@ -1,23 +1,44 @@
...
+    using System.Collections.Generic;
     using System.Threading.Tasks;
     using Gateway.Routing;
     using Pipeline;
     using Routing;
 
     class RouteToGatewayBehavior : Behavior<IRoutingContext>
     {
-        public RouteToGatewayBehavior(string gatewayAddress)
+        public RouteToGatewayBehavior(string gatewayAddress, IList<string> configuredSiteKeys)
```
bording: Why `IList` here?

src/NServiceBus.Gateway/Routing/RouteToGatewayBehaviour.cs
```diff
@@ -1,23 +1,44 @@
...
         }
 
         public override Task Invoke(IRoutingContext context, Func<Task> next)
         {
-
             if (context.Extensions.TryGet(out RouteThroughGateway _))
             {
+                if (context.Message.Headers.TryGetValue(Headers.DestinationSites, out var siteKeyHeader))
+                {
+                    var siteKeys = siteKeyHeader.Split(',');
```
bording: It makes me sad that we take these sites in as a collection, join them to save them, and then split them back out again. Seems like we could do better here and just save them as a collection?

src/NServiceBus.Gateway/Routing/RouteToGatewayBehaviour.cs
```diff
@@ -1,23 +1,44 @@
...
+                    {
+                        if (!configuredSiteKeys.Contains(siteKey))
+                        {
+                            unknownSites.Add(siteKey);
+                        }
+                    }
+
+                    if (unknownSites.Count > 0)
+                    {
+                        throw new Exception($"Sites with keys `{string.Join(",", unknownSites)}` was not found in the list of configured sites. Please make sure to configure it or remove it from the call to `{nameof(SendOptionsExtensions.RouteToSites)}`");
```
bording: This exception message needs some work. Are "keys" even something someone using the Gateway would know about?

They're also much more likely to have used `SendToSites`, so talking about `SendOptionsExtensions.RouteToSites` here seems like it would be confusing. Do we even need that sentence?

Also, there's a grammar problem. Since "Sites" is plural, it should be "were" and not "was".

So, maybe something like:

"The following sites have not been configured: `{string.Join(",", unknownSites)}`"

src/NServiceBus.Gateway.AcceptanceTests/When_sending_a_message_via_the_gateway.cs
```diff
@@ -18,9 +18,8 @@ public async Task Should_process_message()
             var context = await Scenario.Define<Context>()
                 .WithEndpoint<Headquarters>(b => b.When(bus =>
                 {
-#pragma warning disable DE0003 // API is deprecated
```
bording: FYI these warnings come from the Microsoft.DotNet.Analyzers.Compatibility package, which does recommend not using `HttpWebRequest` : https://github.com/dotnet/platform-compat/blob/master/docs/DE0003.md

However, it doesn't look like we've added that package to this repo yet, so we don't have this warning yet.

src/NServiceBus.Gateway/Routing/RouteToGatewayBehaviour.cs
andreasohlund: Talking to @timbussmann we both agree that this is a valid change but not in scope for the patch since its more of a refactoring. I will raise an issue to track this and flag it as a candidate for a maintenance release

src/NServiceBus.Gateway.AcceptanceTests/When_sending_a_message_via_the_gateway.cs
timbussmann: should we switch to start using `HttpClient` in the gateway then? Sounds like something to consider for a maintenance release?

src/NServiceBus.Gateway/Routing/RouteToGatewayBehaviour.cs
andreasohlund: Raise https://github.com/Particular/NServiceBus.Gateway/issues/106

src/NServiceBus.Gateway.AcceptanceTests/When_sending_a_message_via_the_gateway.cs
andreasohlund: Raise https://github.com/Particular/NServiceBus.Gateway/issues/105

src/NServiceBus.Gateway.AcceptanceTests/When_sending_to_invalid_site.cs
```diff
@@ -0,0 +1,42 @@
+﻿namespace NServiceBus.Gateway.AcceptanceTests
+{
+    using System;
+    using AcceptanceTesting;
+    using NUnit.Framework;
+
+    public class When_sending_to_invalid_site : NServiceBusAcceptanceTest
```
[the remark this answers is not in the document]
andreasohlund: Fixed

src/NServiceBus.Gateway/Routing/RouteToGatewayBehaviour.cs
```diff
@@ -1,23 +1,53 @@
 ﻿namespace NServiceBus
 {
     using System;
+    using System.Collections.Generic;
     using System.Threading.Tasks;
     using Gateway.Routing;
     using Pipeline;
     using Routing;
 
     class RouteToGatewayBehavior : Behavior<IRoutingContext>
```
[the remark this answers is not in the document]
andreasohlund: Fixed

PR #120, opened 2019-02-14
Add obsoletes

timbussmann:
Added obsoletes as discussed.

Note that I did only add the obsoletes to the top-level config section element as this is the same approach we took in v6.


bording:
I guess once we're in v4, it should be enough to have stub classes that no longer inherit from the actual problematic types, and we could then have those stubs be in both frameworks, which would let us unify things in v4. Otherwise we'd have to wait for v5... :cry:

PR #133, opened 2019-11-28
Introduce new gateway persistence seam

timbussmann:
Introduces a first proposal of the gateway persistence seam which is moved from the core to the gateway package. There are significant changes from the [interface defined in core](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Gateway/IDeduplicateMessages.cs):
* Different name to avoid confusion between the two different versions of the gateway seam
* Moved to the NServiceBus main namespace (no strong opinions on that but Gateway is already part of the interface name so not sure why it would have to be in it's dedicated namespace?)
* In order to allow persistence options which do not support DTC, the single `DeduplicateMessage` operation which checked + marked the given message id in one go, has been replaced with two corresponding methods instead. This allows the flow to first check for duplicates and mark the message id as known once it has been dispatched to the transport. This ensures the message isn't considered as "known" before it has been dispatched to the transport.
* Removed the `DateTime` parameter as the gateway would always just pass `DateTime.UtcNow` and apparently no implementation is currently making any use of it. I think this can be handled inside the persistence instead if needed.
* `clientId` parameter has been renamed to `messageId` because from all what I can see, that is the value currently passed and clientId is highly confusing. I expect a client id to remain stable.


src/NServiceBus.Gateway/IGatewayDeduplicationPersistence.cs
```diff
@@ -0,0 +1,30 @@
...
+
+    /// <summary>
+    /// Defines the api to provide storage for the gateway deduplication mechanism.
+    /// </summary>
+    public interface IGatewayDeduplicationStorage
+    {
+        /// <summary>
+        /// Returns whether the storage can enlist in a distributed transaction via ambient <see cref="TransactionScope"/> to support exactly-once dispatching.
+        /// </summary>
+        bool SupportsDistributedTransactions { get; }
```
ramonsmits: Why is this needed? What logic would be different knowing the persister supports enlisting? In both cases, both methods would still be invoked.

ramonsmits:
+1 one for creating 2 methods. Just wondering on the implementation. If the persister supports transactions calling `IsDuplicate` should still to an `INSERT` basically MarkAsDispatched would only commit the transaction or be a NOOP if the transaction is outside the control of the persister.

src/NServiceBus.Gateway/IGatewayDeduplicationPersistence.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        /// <summary>
+        /// Returns if the message is a duplicate.
```
ramonsmits:
```suggestion
        /// Returns if the message is a duplicate. Should INSERT in TX context to optimize for happy path.
```

src/NServiceBus.Gateway/IGatewayDeduplicationPersistence.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// </summary>
+        /// <returns>
+        /// <code>true</code> if the message has been receied successfully before and is considered a duplicate. <code>false</code> otherwise.
```
ramonsmits:
```suggestion
        /// <code>true</code> if the message has been received successfully before and is considered a duplicate. <code>false</code> otherwise.
```

src/NServiceBus.Gateway.Tests/ApprovalFiles/APIApprovals.Approve.netframework.approved.txt
```diff
@@ -162,6 +162,12 @@ namespace NServiceBus
         public void Retries(int numberOfRetries, System.TimeSpan timeIncrease) { }
         public void TransactionTimeout(System.TimeSpan timeout) { }
     }
+    public interface IGatewayDeduplicationStorage
```
andreasohlund: Should be NServiceBus.Gateway according to https://github.com/Particular/PlatformArchitecture/blob/master/documentation/cross-cutting-concepts.md#nservicebus-rules

timbussmann: I thought we had some documentation about this but didn't find it at first glance 👍

src/NServiceBus.Gateway/IGatewayDeduplicationPersistence.cs
timbussmann: That's a very implementation specifc comment which shouldn't be here imo.

src/NServiceBus.Gateway/IGatewayDeduplicationPersistence.cs
timbussmann: we discussed this on a call yesterday and concluded that we probably need this for now for correct handling of transaction scopes (e.g. ensuring that there is also no transaction scope when the persistence doesn't support it)

timbussmann:
merging this for now to unblock other PRs.

PR #134, opened 2019-11-29
Use new seam internally

timbussmann:
Based on #133 , this PR changes the internal code to use the new seam API internally. It does this by wrapping core API with the new seam.

Additionally, it adds additional logic to only open a `TransactionScope` if the transport & persistence support it.

Note: this PR does not allow to provide a custom implementation of the new seam yet. That's a task for another PR.

### DIFF
--- a/src/NServiceBus.Gateway/Deduplication/LegacyDeduplicationWrapper.cs
+++ b/src/NServiceBus.Gateway/Deduplication/LegacyDeduplicationWrapper.cs
@@ -1,0 +1,30 @@
+﻿namespace NServiceBus.Gateway.Deduplication
+{
+    using System;
+    using System.Threading.Tasks;
+    using Extensibility;
+
+    class LegacyDeduplicationWrapper : IGatewayDeduplicationStorage
+    {
+        IDeduplicateMessages legacyPersister;
+
+        public LegacyDeduplicationWrapper(IDeduplicateMessages legacyPersister)
+        {
+            this.legacyPersister = legacyPersister;
+        }
+
+        public bool SupportsDistributedTransactions { get; } = true;
+
+        public async Task<bool> IsDuplicate(string messageId, ContextBag context)
+        {
+            var isNewMessage = await legacyPersister.DeduplicateMessage(messageId, DateTime.UtcNow, context)
+                .ConfigureAwait(false);
+            return !isNewMessage;
+        }
+
+        public Task MarkAsDispatched(string messageId, ContextBag context)
+        {
+            return Task.FromResult(0);
+        }
+    }
+}
--- a/src/NServiceBus.Gateway/Deduplication/LegacyDeduplicationWrapper.cs
+++ b/src/NServiceBus.Gateway/Deduplication/LegacyDeduplicationWrapper.cs
@@ -1,13 +1,12 @@
-﻿namespace NServiceBus.Gateway.Deduplication
+﻿namespace NServiceBus.Gateway
 {
     using System;
     using System.Threading.Tasks;
+    using Deduplication;
     using Extensibility;
 
     class LegacyDeduplicationWrapper : IGatewayDeduplicationStorage
     {
-        IDeduplicateMessages legacyPersister;
-
         public LegacyDeduplicationWrapper(IDeduplicateMessages legacyPersister)
         {
             this.legacyPersister = legacyPersister;
@@ -26,5 +25,7 @@
         {
             return Task.FromResult(0);
         }
+
+        IDeduplicateMessages legacyPersister;
     }
 }
--- a/src/NServiceBus.Gateway/Gateway.cs
+++ b/src/NServiceBus.Gateway/Gateway.cs
@@ -89,7 +89,15 @@
             context.Pipeline.Register("GatewayIncomingBehavior", new GatewayIncomingBehavior(), "Extracts gateway related information from the incoming message");
             context.Pipeline.Register("GatewayOutgoingBehavior", new GatewayOutgoingBehavior(), "Puts gateway related information on the headers of outgoing messages");
 
-            context.RegisterStartupTask(b => new GatewayReceiverStartupTask(channelManager, channelReceiverFactory, GetEndpointRouter(context), b.Build<IDispatchMessages>(), b.Build<IDeduplicateMessages>(), b.BuildAll<IDataBus>()?.FirstOrDefault(), gatewayInputAddress));
+            context.RegisterStartupTask(b => new GatewayReceiverStartupTask(
+                channelManager, 
+                channelReceiverFactory, 
+                GetEndpointRouter(context),
+                b.Build<IDispatchMessages>(), 
+                new LegacyDeduplicationWrapper(b.Build<IDeduplicateMessages>()), 
+                b.BuildAll<IDataBus>()?.FirstOrDefault(), 
+                gatewayInputAddress,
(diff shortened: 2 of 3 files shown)


src/NServiceBus.Gateway/Deduplication/LegacyDeduplicationWrapper.cs
```diff
@@ -0,0 +1,30 @@
+﻿namespace NServiceBus.Gateway.Deduplication
```
andreasohlund: Nit: Should be `NServiceBus.Gateway` according to https://github.com/Particular/PlatformArchitecture/blob/master/documentation/cross-cutting-concepts.md#nservicebus-rules ?

src/NServiceBus.Gateway/Deduplication/LegacyDeduplicationWrapper.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    using System;
+    using System.Threading.Tasks;
+    using Extensibility;
+
+    class LegacyDeduplicationWrapper : IGatewayDeduplicationStorage
+    {
+        IDeduplicateMessages legacyPersister;
```
andreasohlund: Nit: Should be a the bottom of the class

src/NServiceBus.Gateway/Gateway.cs
```diff
@@ -89,7 +89,15 @@ protected override void Setup(FeatureConfigurationContext context)
...
-            context.RegisterStartupTask(b => new GatewayReceiverStartupTask(channelManager, channelReceiverFactory, GetEndpointRouter(context), b.Build<IDispatchMessages>(), b.Build<IDeduplicateMessages>(), b.BuildAll<IDataBus>()?.FirstOrDefault(), gatewayInputAddress));
+            context.RegisterStartupTask(b => new GatewayReceiverStartupTask(
+                channelManager, 
+                channelReceiverFactory, 
+                GetEndpointRouter(context),
+                b.Build<IDispatchMessages>(), 
+                new LegacyDeduplicationWrapper(b.Build<IDeduplicateMessages>()), 
+                b.BuildAll<IDataBus>()?.FirstOrDefault(), 
+                gatewayInputAddress,
+                context.Settings.Get<TransportInfrastructure>().TransactionMode));
```
andreasohlund: Should this be using https://github.com/Particular/NServiceBus/blob/ef1bcfd9cc1e9f0fa77c4e1ce499f3bce4a457b3/src/NServiceBus.Core/Transports/TransactionModeSettingsExtensions.cs instead?

src/NServiceBus.Gateway/Receiving/SingleCallChannelReceiver.cs
```diff
@@ -93,13 +108,32 @@ async Task HandleSubmit(CallInfo callInfo)
...
+                        // swallow exception in non-dtc modes.
+                        // When using no transactions, the message has been sent to the transport already. Throwing would cause the operation to be retried and a guaranteed duplicate to be created. By swallowing the exception, the duplicate is only created if the same message is sent to the gateway for another reason.
+                        // When using distributed transactions, throw so that both persistence and transport can rollback atomically.
+                        if (useTransactionScope)
+                        {
+                            throw;
+                        }
+                        else
+                        {
+                            Logger.Warn($"Failed to mark message with id '{callInfo.ClientId}' as dispatched. This message won't be deduplicated.'", e);
```
andreasohlund: Can we add some more details as to why the message couldn't be dispatched? 

>.. since the transport and persister combination doesn't support distributed transactions

or something similar?

src/NServiceBus.Gateway/Gateway.cs
timbussmann: I decided not to use it for two reasons:
* It throws an exception on send-only endpoints and I'd have to catch that
* It uses the actually configured transaction mode, which is something we should do indeed but not what the gateway is doing currently. I'd like to handle that case in a dedicated PR.

src/NServiceBus.Gateway/Receiving/SingleCallChannelReceiver.cs
timbussmann: the exception will hold more information about the issue. The message has actually been dispatched. Given that you can't use DTC with the combination in use, I'm not sure about the added value of mentioning DTC?

src/NServiceBus.Gateway/Gateway.cs
andreasohlund: Got it

src/NServiceBus.Gateway/Deduplication/LegacyDeduplicationWrapper.cs
```diff
@@ -0,0 +1,31 @@
+﻿namespace NServiceBus.Gateway
+{
+    using System;
+    using System.Threading.Tasks;
+    using Deduplication;
+    using Extensibility;
+
+    class LegacyDeduplicationWrapper : IGatewayDeduplicationStorage
```
ramonsmits: Why not just drop support for the old interface? It gets moved to the gateway project anyway and is a breaking change which means recompiling seams.

src/NServiceBus.Gateway/Receiving/SingleCallChannelReceiver.cs
ramonsmits: Why is this called ClientId and not just MessageIdentifier or MessageID?

timbussmann: I have no idea where this naming is coming from. From my understanding it's just the original message id from the sending site but no idea why this is named ClientId. @andreasohlund @janovesk any ideas?

src/NServiceBus.Gateway/Deduplication/LegacyDeduplicationWrapper.cs
timbussmann: the plan is to release a new minor of the gateway, so that's not the time for breaking changes.

src/NServiceBus.Gateway/Receiving/SingleCallChannelReceiver.cs
andreasohlund: It's just always been that way. My hunch is that its due to the fact that we allow webclients to send raw messages via the gateway

https://github.com/Particular/docs.particular.net/blob/master/samples/gateway/Gateway_3/WebClient/Index.html#L16

so that might be where this comes from

timbussmann: client id would still be highly misleading in that case as a client would probably not change it's "client id" for every message it dispatches.

andreasohlund:
Agreed

src/NServiceBus.Gateway/Receiving/SingleCallChannelReceiver.cs
andreasohlund: agreed

src/NServiceBus.Gateway/Deduplication/LegacyDeduplicationWrapper.cs
ramonsmits: My bad @timbussmann I thought this `IGatewayDeduplicationStorage` was cloned to the gateway but this is the interface defined in the core assembly.

PR #135, opened 2019-12-02
Persistence seam tests

timbussmann:
Adds a set of tests which can be pulled in as a source-only package on downstream implementations to verify the behavior of the implementation


src/NServiceBus.Gateway.PersistenceTests/When_supporting_distributed_transactions.cs
```diff
@@ -0,0 +1,82 @@
...
+    using NUnit.Framework;
+
+    class When_supporting_distributed_transactions : NServiceBusGatewayDeduplicationStorageTest
+    {
+        [SetUp]
+        public void IgnoreNonDTCStorages()
+        {
+            if (!storage.SupportsDistributedTransactions)
+            {
+                Assert.Ignore();
```
ramonsmits: Nice trick to do this in the setup!

timbussmann:
rebased and updated, this needs to be re-approved

src/NServiceBus.Gateway.PersistenceTests/NServiceBus.Gateway.PersistenceTests.csproj
```diff
@@ -0,0 +1,39 @@
...
+  <PropertyGroup>
+    <PackageId>NServiceBus.Gateway.PersistenceTests.Sources</PackageId>
+    <Description>Tests for gateway deduplication storage seam implementations</Description>
+    <IsPackable>true</IsPackable>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" PrivateAssets="All" />
+    <PackageReference Include="NUnit3TestAdapter" Version="3.12.0" PrivateAssets="All" />
```
danielmarbach: If we ship this as a package would it make sense to make this a range like

```
<PackageReference Include="NUnit" Version="[3.12.0, 4.0.0)" />
```

like see https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.TransportTests/NServiceBus.TransportTests.csproj#L18

src/NServiceBus.Gateway.PersistenceTests/NServiceBusGatewayDeduplicationStorageTest.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace NServiceBus.Gateway.PersistenceTests
+{
+    using NUnit.Framework;
+
+    [TestFixture]
+    abstract class NServiceBusGatewayDeduplicationStorageTest
```
danielmarbach: Aren't we usually making those classes public?

Example

https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.TransportTests/NServiceBusTransportTest.cs#L17
https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.AcceptanceTests/NServiceBusAcceptanceTest.cs#L13

src/NServiceBus.Gateway.PersistenceTests/When_checking_unknown_message.cs
```diff
@@ -0,0 +1,35 @@
+﻿namespace NServiceBus.Gateway.PersistenceTests
+{
+    using System;
+    using System.Threading.Tasks;
+    using Extensibility;
+    using NUnit.Framework;
+
+    class When_checking_unknown_message : NServiceBusGatewayDeduplicationStorageTest
```
danielmarbach: Same question why not public?

src/NServiceBus.Gateway.PersistenceTests/When_supporting_distributed_transactions.cs
```diff
@@ -0,0 +1,112 @@
...
+            }
+
+            using (var session = await storage.CheckForDuplicate(messageId, new ContextBag()))
+            {
+                Assert.IsTrue(session.IsDuplicate);
+            }
+        }
+
+        [Test]
+        public async Task Does_not_mark_messages_as_dispatched_when_aborted()
```
danielmarbach: `when_rolled_back` maybe slightly better?

src/NServiceBus.Gateway.PersistenceTests/When_supporting_distributed_transactions.cs
```diff
@@ -0,0 +1,112 @@
...
+        [Test]
+        public async Task Only_marks_messages_as_dispatched_when_committed()
+        {
+            var messageId = Guid.NewGuid().ToString("D");
+
+            using (var scope = new TransactionScope(TransactionScopeOption.RequiresNew, TransactionScopeAsyncFlowOption.Enabled))
+            {
+                using (var session = await storage.CheckForDuplicate(messageId, new ContextBag()))
+                {
+                    Assert.IsFalse(session.IsDuplicate);
```
danielmarbach: I'm not such a huge fan of mixing arrange and asserts. I can so though how this can work here. Another option would be to capture flags and then Assert at the end which would be my preference but I'm not hung up on it

src/NServiceBus.Gateway.PersistenceTests/When_marking_message_as_dispatched.cs
```diff
@@ -0,0 +1,43 @@
...
+            }
+
+            using (var session = await storage.CheckForDuplicate(messageId, new ContextBag()))
+            {
+                Assert.IsTrue(session.IsDuplicate);
+            }
+        }
+
+        [Test]
+        public async Task MarkAsDispatched_can_be_called_multiple_times()
```
danielmarbach: I would wrap the whole block in an Assert.DoesNotThrow. I know that if it throws the test fails as well but I think it is more expressive to express the assumptions with Asserts

src/NServiceBus.Gateway.PersistenceTests/NServiceBus.Gateway.PersistenceTests.csproj
timbussmann: this has PrivateAssets=All so it won't be listed as a dependency

src/NServiceBus.Gateway.PersistenceTests/When_supporting_distributed_transactions.cs
timbussmann: I have no strong preference either.

src/NServiceBus.Gateway.PersistenceTests/When_marking_message_as_dispatched.cs
timbussmann: 👍

src/NServiceBus.Gateway.PersistenceTests/When_checking_unknown_message.cs
timbussmann: 🤷‍♂ didn't see a need for it to be public. No strong opinion though.

src/NServiceBus.Gateway.PersistenceTests/NServiceBus.Gateway.PersistenceTests.csproj
```diff
@@ -0,0 +1,39 @@
...
+    <ProjectReference Include="..\NServiceBus.Gateway\NServiceBus.Gateway.csproj" />
+  </ItemGroup>
+
+  <!-- Workaround for https://github.com/dotnet/sdk/issues/1479 -->
+  <ItemGroup>
+    <_PackageFiles Include="**\*.cs">
+      <BuildAction>Compile</BuildAction>
+      <PackagePath>content\App_Packages\NSB.TransportTests\;contentFiles\cs\any\NSB.TransportTests\</PackagePath>
+    </_PackageFiles>
+    <_PackageFiles Remove="**\obj\**\*.cs" />
```
ramonsmits: Why only obj? Isn't there a bin folder present as well?

src/NServiceBus.Gateway.PersistenceTests/When_marking_message_as_dispatched.cs
ramonsmits: or, explicitly do `Assert.Pass();` at the end as an alternative. I personally like the test already as it is.

src/NServiceBus.Gateway.PersistenceTests/When_supporting_distributed_transactions.cs
ramonsmits: I like KISS and less is more so I would choose the current form.

src/NServiceBus.Gateway.PersistenceTests/NServiceBus.Gateway.PersistenceTests.csproj
danielmarbach: Shouldn't this enforce NUnit given that it is required and therefore not have private assets?

src/NServiceBus.Gateway.PersistenceTests/NServiceBus.Gateway.PersistenceTests.csproj
timbussmann: no bin files are already filtered out. A few lines above it only picks `*.cs` files but there can be some in the obj folder we don't want.

src/NServiceBus.Gateway.PersistenceTests/NServiceBus.Gateway.PersistenceTests.csproj
timbussmann: I think this is how core does it, I guess the adapter is required to make the tests actually run but not for compilation. I'm not familiar with the test adapter package release strategy tbh so not sure it makes sense to change this atm.

danielmarbach: I'm talking about NUnit.

danielmarbach:
Sent in https://github.com/Particular/NServiceBus.Gateway/pull/144

PR #137, opened 2019-12-03
Add configuration API for the new persistence seam

timbussmann:
Adds a new API `IGatewayDeduplicationConfiguration` which allows to define and configure the deduplication storage without actually building the storage already. This an approach to allow the storage to be built only at endpoint startup time and even make use of the builder if needed (e.g. allowing it to resolve sessions from a container)

@andreasohlund we talked about this in regards to some persistence seam where this was a limitation to the RavenDB persistence. What do you think about this approach?


src/NServiceBus.Gateway/Configuration/IGatewayDeduplicationConfiguration.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace NServiceBus.Gateway
+{
+    using ObjectBuilder;
+
+    /// <summary>
+    /// Configures the deduplication storage.
+    /// </summary>
+    public interface IGatewayDeduplicationConfiguration
```
andreasohlund: Is there a benefit of this being an `interface`? Thinking that if this was an `abstract class` it would be easier to evolve WO breaking changes?

timbussmann: no specific reason tbh. I was thinking about an abstract class as well but couldn't find a clear advantage. Maybe the reason you stated might be good enough, happy to change it as I don't prefer a specific approach and the abstract class is probably closer to the core's API.

andreasohlund: 👍 for me being able to evolve without breaking SemVer is nice advantage

timbussmann: changed the configuration API to an abstract class, have another look.

PR #138, opened 2019-12-05
Session based API

timbussmann:
an attempt to design a different gateway API to allow persisters to make use of transactions by introducing a session that links the `IsDuplicate` check with the `MarkAsDispatched` operation.

the current API approach looks roughly like this:
```c#
if(!storage.IsDuplicate(messageId))
{
    transport.Dispatch(message);
    storage.MarkAsDispatched(messageID);
}
```
this makes it difficult for persistence to make use of session/transaction based functionalities like "select for update" as it's not really possible to open and share a transaction between these storage operations.

This approach might solve this problem:

```c#
using(var session = storage.IsDuplicate(message))
{
   if(!session.IsDuplicate)
   {
        transport.Dispatch(message);
        session.MarkAsDispatched();
   }
}
```

the session can encapsulate a transaction and share it easily with the `MarkAsDispatched` operation while having a clear lifetime due to the usage of `IDisposable`

### DIFF
--- a/src/NServiceBus.Gateway.Tests/ApprovalFiles/APIApprovals.Approve.netframework.approved.txt
+++ b/src/NServiceBus.Gateway.Tests/ApprovalFiles/APIApprovals.Approve.netframework.approved.txt
@@ -112,7 +112,7 @@ namespace NServiceBus.Gateway
     {
         System.Threading.Tasks.Task Send(string remoteAddress, System.Collections.Generic.IDictionary<string, string> headers, System.IO.Stream data);
     }
-    public interface IDuplicationCheckSession : System.IDisposable
+    public interface IDeduplicationSession : System.IDisposable
     {
         bool IsDuplicate { get; }
         System.Threading.Tasks.Task MarkAsDispatched();
@@ -120,7 +120,7 @@ namespace NServiceBus.Gateway
     public interface IGatewayDeduplicationStorage
     {
         bool SupportsDistributedTransactions { get; }
-        System.Threading.Tasks.Task<NServiceBus.Gateway.IDuplicationCheckSession> CheckForDuplicate(string messageId, NServiceBus.Extensibility.ContextBag context);
+        System.Threading.Tasks.Task<NServiceBus.Gateway.IDeduplicationSession> CheckForDuplicate(string messageId, NServiceBus.Extensibility.ContextBag context);
     }
     public class InMemoryDeduplicationConfiguration : NServiceBus.Gateway.GatewayDeduplicationConfiguration
     {
--- a/src/NServiceBus.Gateway.Tests/ApprovalFiles/APIApprovals.Approve.netframework.approved.txt
+++ b/src/NServiceBus.Gateway.Tests/ApprovalFiles/APIApprovals.Approve.netframework.approved.txt
@@ -112,11 +112,15 @@ namespace NServiceBus.Gateway
     {
         System.Threading.Tasks.Task Send(string remoteAddress, System.Collections.Generic.IDictionary<string, string> headers, System.IO.Stream data);
     }
+    public interface IDuplicationCheckSession : System.IDisposable
+    {
+        bool IsDuplicate { get; }
+        System.Threading.Tasks.Task MarkAsDispatched();
+    }
     public interface IGatewayDeduplicationStorage
     {
         bool SupportsDistributedTransactions { get; }
-        System.Threading.Tasks.Task<bool> IsDuplicate(string messageId, NServiceBus.Extensibility.ContextBag context);
-        System.Threading.Tasks.Task MarkAsDispatched(string messageId, NServiceBus.Extensibility.ContextBag context);
+        System.Threading.Tasks.Task<NServiceBus.Gateway.IDuplicationCheckSession> CheckForDuplicate(string messageId, NServiceBus.Extensibility.ContextBag context);
     }
     public class InMemoryDeduplicationConfiguration : NServiceBus.Gateway.GatewayDeduplicationConfiguration
     {
--- a/src/NServiceBus.Gateway.Tests/ApprovalFiles/APIApprovals.Approve.netstandard.approved.txt
+++ b/src/NServiceBus.Gateway.Tests/ApprovalFiles/APIApprovals.Approve.netstandard.approved.txt
@@ -38,11 +38,15 @@ namespace NServiceBus.Gateway
     {
         System.Threading.Tasks.Task Send(string remoteAddress, System.Collections.Generic.IDictionary<string, string> headers, System.IO.Stream data);
     }
+    public interface IDuplicationCheckSession : System.IDisposable
+    {
+        bool IsDuplicate { get; }
+        System.Threading.Tasks.Task MarkAsDispatched();
+    }
     public interface IGatewayDeduplicationStorage
     {
         bool SupportsDistributedTransactions { get; }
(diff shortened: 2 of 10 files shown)


andreasohlund:
Lgtm

src/NServiceBus.Gateway/Deduplication/LegacyDeduplicationWrapper.cs
```diff
@@ -14,18 +14,32 @@ public LegacyDeduplicationWrapper(IDeduplicateMessages legacyPersister)
 
         public bool SupportsDistributedTransactions { get; } = true;
 
-        public async Task<bool> IsDuplicate(string messageId, ContextBag context)
+        public async Task<IDuplicationCheckSession> IsDuplicate(string messageId, ContextBag context)
```
ramonsmits: The method should be something else like `CreateDuplicateSessionForMessageId` or something.

src/NServiceBus.Gateway/InMemoryStorage/InMemoryDeduplicationStorage.cs
```diff
@@ -50,6 +50,29 @@ public Task MarkAsDispatched(string messageId, ContextBag context)
 
                 return Task.FromResult(0);
             }
+
+            public void Dispose()
+            {
+                Monitor.Exit(lockObj);
```
ramonsmits:
```suggestion
                Monitor.Exit(lockObj);
                GC.SuppressFinalize(this);                
```

ramonsmits:
@timbussmann I like the idea!

src/NServiceBus.Gateway/Deduplication/LegacyDeduplicationWrapper.cs
timbussmann: yeah I didn't spend time to rethink the method names on the spike. I agree that we should change this name though 👍

timbussmann:
should be good for final reviews.

I've changed the locking on the inmemory persistence so that not the whole session is locked but only the collection access is locked individually. This means:
When configuring higher concurrency levels than one, the chance of duplicates can increase as sessions aren't executed sequentially (due to blocking locking) but concurrently. Therefore concurrently processing multiple requests for the same message can each return false for the `IsDuplicate` check. I think that's desired behavior as there is a reason the concurrency level has been increased by the user and it would have no impact if the whole duplication check uses a single blocking lock.

src/NServiceBus.Gateway.Tests/ApprovalFiles/APIApprovals.Approve.netstandard.approved.txt
```diff
@@ -38,11 +38,15 @@ namespace NServiceBus.Gateway
...
         System.Threading.Tasks.Task Send(string remoteAddress, System.Collections.Generic.IDictionary<string, string> headers, System.IO.Stream data);
     }
+    public interface IDuplicationCheckSession : System.IDisposable
+    {
+        bool IsDuplicate { get; }
+        System.Threading.Tasks.Task MarkAsDispatched();
+    }
     public interface IGatewayDeduplicationStorage
     {
         bool SupportsDistributedTransactions { get; }
```
ramonsmits: Rename to `SupportsTransactionScope`?

src/NServiceBus.Gateway.Tests/ApprovalFiles/APIApprovals.Approve.netframework.approved.txt
```diff
@@ -112,11 +112,15 @@ namespace NServiceBus.Gateway
...
+        System.Threading.Tasks.Task MarkAsDispatched();
+    }
     public interface IGatewayDeduplicationStorage
     {
         bool SupportsDistributedTransactions { get; }
-        System.Threading.Tasks.Task<bool> IsDuplicate(string messageId, NServiceBus.Extensibility.ContextBag context);
-        System.Threading.Tasks.Task MarkAsDispatched(string messageId, NServiceBus.Extensibility.ContextBag context);
+        System.Threading.Tasks.Task<NServiceBus.Gateway.IDuplicationCheckSession> CheckForDuplicate(string messageId, NServiceBus.Extensibility.ContextBag context);
     }
     public class InMemoryDeduplicationConfiguration : NServiceBus.Gateway.GatewayDeduplicationConfiguration
```
ramonsmits: Maybe we should also rename `GatewayDeduplicationConfiguration` to `GatewayDeduplicationStorageConfiguration`?

src/NServiceBus.Gateway/InMemoryStorage/InMemoryDeduplicationSession.cs
```diff
@@ -1,57 +1,65 @@
...
-            GC.SuppressFinalize(this);
         }
 
         readonly string messageId;
         readonly Dictionary<string, LinkedListNode<string>> clientIdSet;
         readonly LinkedList<string> clientIdList;
-        readonly object lockObj;
         readonly int cacheSize;
+
+        static readonly object lockObj = new object();
```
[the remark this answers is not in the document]
timbussmann: this class isn't a single instance. Every check has it's dedicated instance as the session is tied to a message id. The collections are shared across those sessions and therefore need a shared lock.
I went with a dedicated lock object as that is considered the best practice.

src/NServiceBus.Gateway.Tests/ApprovalFiles/APIApprovals.Approve.netstandard.approved.txt
timbussmann: I think that wouldn't capture the intention  correctly. While a transaction scope is used, we only use a transaction scope when a distributed transaction should be used. What if a storage could enlist with a transaction scope but not with distributed transactions?

src/NServiceBus.Gateway.Tests/ApprovalFiles/APIApprovals.Approve.netframework.approved.txt
timbussmann: I tried to keep the name as short as possible. I didn't feel the "Storage" part would add anything meaningful to the name?

ramonsmits: Not a lot, just that this is specific to storage. Its not a big thing.

src/NServiceBus.Gateway.Tests/ApprovalFiles/APIApprovals.Approve.netstandard.approved.txt
ramonsmits: I'm not sure if that would change the outcome of the behavior. I don't know the ins/ours ot transaction scope implementation behavior principles.

src/NServiceBus.Gateway/InMemoryStorage/InMemoryDeduplicationSession.cs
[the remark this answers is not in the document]
ramonsmits: Isn't this than done on the reference so achieve the same outcome? Hosting multiple endpoints in a single process would now share the same lockobj but have their own isolated state.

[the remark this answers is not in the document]
timbussmann: I see what you mean. Yes, in case of a multi-endpoint hosting scenario, all in-mem storages would shame the same lock. Good point, I can change that.

PR #139, opened 2019-12-11
Use configured transport transaction mode

timbussmann:
Uses the configured transport transaction mode to determine whether to wrap the gateway operations in a `TransactionScope` or not (instead of the transports highest supported mode).
This allows to use DTC supporting transports to disable the `TransactionScope` in cases they do not want to make use of DTC.

Note that this only works as long as the endpoint hosting the gateway isn't configured as a send-only endpoint. When using a send-only endpoint, the transport's highest supported transaction mode is used for the decision to use `TransactionScope`.

### DIFF
--- a/src/NServiceBus.Gateway/Gateway.cs
+++ b/src/NServiceBus.Gateway/Gateway.cs
@@ -4,6 +4,7 @@
     using System.Collections.Generic;
     using System.Linq;
     using System.Threading.Tasks;
+    using ConsistencyGuarantees;
     using DeliveryConstraints;
     using Extensibility;
     using Logging;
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Gateway/Gateway.cs
```diff
@@ -91,14 +92,28 @@ protected override void Setup(FeatureConfigurationContext context)
...
+        {
+            try
+            {
+                return context.Settings.GetRequiredTransactionModeForReceives();
+            }
+            catch (Exception)
+            {
+                // GetRequiredTransactionModeForReceives throws on read-only endpoints.
+                // Use the transport's default mode in that case:
+                return context.Settings.Get<TransportInfrastructure>().TransactionMode;
```
andreasohlund: Do we have a core test to document this "backdoor"? (If not, should we add one?)

timbussmann: I can check this. I think we do but I'll add a test otherwise.

timbussmann: PR created here: https://github.com/Particular/NServiceBus/pull/5542

PR #141, opened 2019-12-11
Use provided configuration

timbussmann:
storage provided using the new `Gateway(storage)` API isn't taken into account right now unless there is also a storage type specified due to the check in the Gateway feature.

This check has been moved to the "legacy" storage implementation (default when using the parameterless `Gateway()` API). This required adding a new method to the configuration class which can be optionally implemented to run checks or setup the persister based on the read-only settings.

Based on that I had to adjust the acceptance tests to the new APIs. Since each test does a lot of gateway specific configuration, they currently have to retrieve the `GatewaySettings` from the settings bag. I'd like to find a better approach for that but it solves the issue for now.


timbussmann:
looks like a permission issue. Not sure why this triggers with these changes. Going to rerun tests before investigating further.

PR #142, opened 2019-12-12
Skip installer when not running on Windows

timbussmann:
When running on linux, the installer seems to usually trip over `Environment.OSVersion.Version.Major <= 5` and not run the installer (which is correct). The behavior does seem to be "lucky coincidence"  and instead we should avoid running the installer logic when not running on Windows.

### DIFF
--- a/src/NServiceBus.Gateway/Installer/GatewayHttpListenerInstaller.cs
+++ b/src/NServiceBus.Gateway/Installer/GatewayHttpListenerInstaller.cs
@@ -34,6 +34,12 @@ namespace NServiceBus.Installation
                 return Task.FromResult(0);
             }
 
+            if (Environment.OSVersion.Platform != PlatformID.Win32NT)
+            {
+                logger.InfoFormat("Installer does not support platform {0}. Ensure that the process has required permissions to use listen to configured urls.", Environment.OSVersion.Platform.ToString("G"));
+                return Task.FromResult(0);
+            }
+
             if (Environment.OSVersion.Version.Major <= 5)
             {
                 logger.InfoFormat(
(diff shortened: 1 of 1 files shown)


timbussmann:
@bording any thoughts on this given that Microsoft recommends to not use `OSVersion` when possible?

src/NServiceBus.Gateway/Installer/GatewayHttpListenerInstaller.cs
```diff
@@ -34,6 +34,12 @@ public Task Install(string identity)
                 return Task.FromResult(0);
             }
 
+            if (Environment.OSVersion.Platform != PlatformID.Win32NT)
```
danielmarbach:
```suggestion
            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```

src/NServiceBus.Gateway/Installer/GatewayHttpListenerInstaller.cs
```diff
@@ -34,6 +34,12 @@ public Task Install(string identity)
...
+            {
+                logger.InfoFormat("Installer does not support platform {0}. Ensure that the process has required permissions to use listen to configured urls.", Environment.OSVersion.Platform.ToString("G"));
```
danielmarbach:
```suggestion
                logger.InfoFormat("Installer does not support platform {0}. Ensure that the process has required permissions to use listen to configured urls.", RuntimeInformation.OSDescription);
```

ramonsmits: I also removed "USE" from the string

PR #143, opened 2019-12-12
Legacy storage detection

timbussmann:
Adds a check to detect if core's in-memory implementation for the gateway storage is being used and logs a warning to use the built-in version of the gateway instead.

### DIFF
--- a/src/NServiceBus.Gateway.AcceptanceTests/When_using_legacy_inmemory_persistence.cs
+++ b/src/NServiceBus.Gateway.AcceptanceTests/When_using_legacy_inmemory_persistence.cs
@@ -1,0 +1,39 @@
+﻿namespace NServiceBus.Gateway.AcceptanceTests
+{
+    using System.Linq;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using Logging;
+    using NUnit.Framework;
+
+    public class When_using_legacy_inmemory_persistence : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_log_warning()
+        {
+            var context = await Scenario.Define<ScenarioContext>()
+                .WithEndpoint<EndpointWithLegacyConfiguration>()
+                .Done(c => c.EndpointsStarted)
+                .Run();
+
+            var warnings = context.Logs
+                .Where(l => l.Level == LogLevel.Warn)
+                .Select(l => l.Message);
+            Assert.That(warnings.Any(l => l.Contains("Endpoint is configured to use the legacy in-memory gateway deduplication storage")));
+        }
+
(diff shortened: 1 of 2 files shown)


danielmarbach:
@timbussmann the test seems to fail on .NET Core on linux because it try to binds something it doesn't have the right to do so. Any thoughts?

src/NServiceBus.Gateway/Deduplication/LegacyDeduplicationStorageConfiguration.cs
```diff
@@ -22,12 +24,19 @@ public override void Setup(ReadOnlySettings settings)
                 throw new Exception("No persistence configured, please configure one that supports gateway deduplication storage or provide a deduplication storage via the 'endpointConfiguration.Gateway' API.");
             }
 
+            if (settings.IsFeatureEnabled(typeof(InMemoryGatewayPersistence)))
+            {
+                log.WarnFormat("Endpoint is configured to use the legacy in-memory gateway deduplication storage. This storage is going to be obsoleted in future versions of NServiceBus. Use the gateway's built-in in-memory storage instead: endpointConfiguration.UseGateway(new {0}());'", nameof(InMemoryDeduplicationConfiguration));
```
ramonsmits:
```suggestion
                log.Warn($"Endpoint is configured to use the legacy in-memory gateway deduplication storage. This storage is going to be obsoleted in future versions of NServiceBus. Use the gateway's built-in in-memory storage instead: endpointConfiguration.UseGateway(new {nameof(InMemoryDeduplicationConfiguration)}());'");
```

This does not require a formatted structured error to be written as the type is static.

src/NServiceBus.Gateway.AcceptanceTests/When_using_legacy_inmemory_persistence.cs
```diff
@@ -31,7 +31,8 @@ public EndpointWithLegacyConfiguration()
                     c.UsePersistence<InMemoryPersistence, StorageType.GatewayDeduplication>();
                     
                     // legacy configuration API (not passing a storage as parameter):
-                    c.Gateway();
+                    var gatewayConfiguration = c.Gateway();
+                    gatewayConfiguration.AddReceiveChannel("http://localhost:25999/SiteA/");
```
ramonsmits: Why was this API not required in the test before?

timbussmann: this is a new test and the test failed as the default receive channel url doesn't seem to be configured on the build-agent. Not really sure how it's done, but the specified URL is used by multiple tests and seems to be allowed on the agent.

PR #146, opened 2019-12-17
TreatAsErrorFromVersion must not be 3.0 as that prevent recompile

ramonsmits:
The gateway type must still be usable, but users should be getting a hint that they should migrate to the other gateway API.

Meaning in v3 a warning, in v4 an error with NotImplementedException and in v5 removed.


timbussmann:
> The gateway type must still be usable, but users should be getting a hint that they should migrate to the other gateway API.

why does it need to be usable? Keep in mind that this is the code which has already been released as part of 3.0

PR #150, opened 2019-12-17
Update Particular.Packaging in release-3.1

DavidBoike:
For your review @bording


bording:
Even though 3.1.0 isn't released yet, since it is already staged, it's safe to go ahead and merge this. When 3.1.0 is pushed to production, the correct commit will be tagged even if this has been merged already.

PR #226, opened 2020-12-10
Fix API Approvals

SeanFeldman:


SeanFeldman:
@bording API Approvals between .NET Framework and .NET Standard are different.

src/NServiceBus.Gateway.Tests/APIApprovals.cs
```diff
@@ -17,7 +17,7 @@ public void Approve()
 #else
             var targetFramework = "netstandard";
 #endif
-            var publicApi = ApiGenerator.GeneratePublicApi(typeof(GatewaySettings).Assembly);
+            var publicApi = ApiGenerator.GeneratePublicApi(typeof(GatewaySettings).Assembly, excludeAttributes: new[] { "System.Runtime.Versioning.TargetFrameworkAttribute", "System.Reflection.AssemblyMetadataAttribute" });
```
bording, line 20: I think I would like to keep the TargetFrameworkAttribute in the files since we are splitting the approval files by target framework.

SeanFeldman: Given it's a release branch, I'd say we keep it as-is.

bording: Keeping it as-is, would be to not add the exclusion. This PR is adding it.

SeanFeldman: Ah, right. Fixed.

bording (approved):

PR #267, opened 2021-04-27
Implement a Gateway compatibility with linux and K8s

joseluis-balsera:
- Enables execution in a docker container or K8s pod.
- Adds the option to disable the wildcard control in K8s pods.

Hosting the gateway in a Kubernetes pod was not possible because the installer tried to add the URL ACL. 
We added the possibility to inform the gateway it is in a Kubernetes environment and then use wildcard binding.

### DIFF
--- a/src/NServiceBus.Gateway/Gateway.cs
+++ b/src/NServiceBus.Gateway/Gateway.cs
@@ -3,6 +3,9 @@
     using System;
     using System.Collections.Generic;
     using System.Linq;
+#if NETSTANDARD
+    using System.Runtime.InteropServices;
+#endif
     using System.Threading.Tasks;
     using ConsistencyGuarantees;
     using DeliveryConstraints;
@@ -117,7 +120,12 @@
                 return;
             }
 
-            CheckForNonWildcardDefaultChannel(channelManager);
+
+            var k8SIsActive = context.Settings.Get<bool>(GatewaySettings.KubernetesCompatibilityIsActive);
+            if (!k8SIsActive)
+            {
+                CheckForNonWildcardDefaultChannel(channelManager);
+            }
 
             channelReceiverFactory = s => new ChannelReceiverFactory(typeof(HttpChannelReceiver)).GetReceiver(s);
             channelSenderFactory = s => new ChannelSenderFactory(typeof(HttpChannelSender)).GetSender(s);
@@ -125,6 +133,12 @@
             var installerSettings = context.Settings.Get<InstallerSettings>();
             installerSettings.ChannelManager = channelManager;
             installerSettings.Enabled = true;
+
+#if NETSTANDARD
+           installerSettings.Enabled = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+#else
+            installerSettings.Enabled = Environment.OSVersion.Platform == PlatformID.Win32NT;
+#endif
         }
 
         static void CheckForNonWildcardDefaultChannel(IManageReceiveChannels channelManager)
@@ -143,7 +157,7 @@
 
         static EndpointRouter GetEndpointRouter(FeatureConfigurationContext context)
         {
-            return new EndpointRouter {MainInputAddress = context.Settings.EndpointName()};
+            return new EndpointRouter { MainInputAddress = context.Settings.EndpointName() };
         }
 
         static void ConfigureTransaction(FeatureConfigurationContext context)
--- a/src/NServiceBus.Gateway/Settings/GatewaySettings.cs
+++ b/src/NServiceBus.Gateway/Settings/GatewaySettings.cs
@@ -18,9 +18,15 @@
     /// </summary>
     public class GatewaySettings
     {
+        /// <summary>
+        /// Settings key to enable/disable K8s compatibility
+        /// </summary>
+        public const string KubernetesCompatibilityIsActive = "KubernetesCompatibilityIsActive";
+
         internal GatewaySettings(EndpointConfiguration config)
         {
             settings = config.GetSettings();
+            settings.Set(KubernetesCompatibilityIsActive, false);
         }
 
 
@@ -40,6 +46,14 @@
 
             settings.Set("GatewayChannelSenderFactory", senderFactory);
             settings.Set("GatewayChannelReceiverFactory", receiverFactory);
+        }
+
+        /// <summary>
+        /// Activate a wildcard on Kubernetes Pods avoid a CheckForNonWildcardDefaultChannel(channelManager) on feature enabling.
+        /// </summary>
+        public void ActivateKubernetesCompatibility()
+        {
+            settings.Set(KubernetesCompatibilityIsActive, true);
         }
 
 
@@ -135,7 +149,7 @@
         {
             Guard.AgainstNegativeAndZero(nameof(timeout), timeout);
 
-            settings.Set("Gateway.TransactionTimeout",timeout);
+            settings.Set("Gateway.TransactionTimeout", timeout);
         }
 
         internal static TimeSpan? GetTransactionTimeout(ReadOnlySettings settings)


DavidBoike:
Hi @joseluis-balsera,

First of all I'm sorry this PR stagnated for so long that Stalebot got involved. We have internal procedures to notify us about external PRs and clearly that broke down.

On the actual PR, you said the installer tries to add the urlacl. I'm not sure I follow, since your code changes aren't touching any installers, but only the Gateway feature. Do you mean that if you specify a non-wildcard address, something happens somewhere else in an installer? Are you talking about [this method](https://github.com/Particular/NServiceBus.Gateway/blob/release-3.1/src/NServiceBus.Gateway/Installer/GatewayHttpListenerInstaller.cs#L33)? And if so, have you tried not enabling installers? I think if that works, it would be preferable.

If it's something else, I'd prefer to not name a feature or API directly after Kubernetes. After all it's not explicitly that it's running in Kubernetes that's the issue, but there's something else going on that happens to be a problem in a K8S environment. That's why I'm trying to do a bit of 5 Whys on you to discover the true root of the issue.

emasdeu:
Hi David,

In fact, we are talking about two different changes in the same PR:

1) Compatibility with linux (our dockers are linux based)
The idea is to disable installers for linux cases.
The "enabled.installers = true", executes "netsh..." which works only in Windows.
That is why we change in Gateway.cs to condition whether it must be enabled or not depending on the operation system


2) Compatibility with Kubernetes
Wildcards (*, +) were disabled. In case of Kubernetes, this was a problem since after the proxy the pod host name is dynamic.
If you prefer, we can change the naming and talk about "ChannelURIEnableWildcards" instead of "ActivateK8Compatibility"

If you prefer, we can make a call to clarify your doubts.

Enric Masdeu



De: David Boike ***@***.***>
Enviado el: martes, 1 de junio de 2021 17:13
Para: Particular/NServiceBus.Gateway ***@***.***>
CC: Subscribed ***@***.***>
Asunto: Re: [Particular/NServiceBus.Gateway] Implement a Gateway compatibility with linux and K8s (#267)

Caution, this email may be from a sender outside Wolters Kluwer. Verify the sender and know the content is safe.

Atención, este correo electrónico proviene de un remitente externo a Wolters Kluwer. Proceda a verificarlo para confirmar que el contenido es seguro.

Hi @joseluis-balsera<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fjoseluis-balsera&data=04%7C01%7Cenric.masdeu%40wolterskluwer.com%7C858e78020e814e409e3e08d9250fbeb0%7C8ac76c91e7f141ffa89c3553b2da2c17%7C0%7C0%7C637581571826625868%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=QO7F4Xho5m8CaTCxU4vLxVoEXRB66Np3tYd2N82UUhQ%3D&reserved=0>,

First of all I'm sorry this PR stagnated for so long that Stalebot got involved. We have internal procedures to notify us about external PRs and clearly that broke down.

On the actual PR, you said the installer tries to add the urlacl. I'm not sure I follow, since your code changes aren't touching any installers, but only the Gateway feature. Do you mean that if you specify a non-wildcard address, something happens somewhere else in an installer? Are you talking about this method<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FParticular%2FNServiceBus.Gateway%2Fblob%2Frelease-3.1%2Fsrc%2FNServiceBus.Gateway%2FInstaller%2FGatewayHttpListenerInstaller.cs%23L33&data=04%7C01%7Cenric.masdeu%40wolterskluwer.com%7C858e78020e814e409e3e08d9250fbeb0%7C8ac76c91e7f141ffa89c3553b2da2c17%7C0%7C0%7C637581571826635861%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=QC5JSf%2BNlbX5vUQM2G8sFaBrnYfFVtC0UjMUVfeKw%2FU%3D&reserved=0>? And if so, have you tried not enabling installers? I think if that works, it would be preferable.

If it's something else, I'd prefer to not name a feature or API directly after Kubernetes. After all it's not explicitly that it's running in Kubernetes that's the issue, but there's something else going on that happens to be a problem in a K8S environment. That's why I'm trying to do a bit of 5 Whys on you to discover the true root of the issue.

-

DavidBoike:
Ok, I get it a little bit more now.

For case 1, I don't think this should be necessary. In [this block of the installer](https://github.com/Particular/NServiceBus.Gateway/blob/release-3.1/src/NServiceBus.Gateway/Installer/GatewayHttpListenerInstaller.cs#L40-L51) we're essentially already doing the same check and performing an early exit if you're not on Windows. Do you see something else happening on your end? Is it possible those conditions need to be tweaked?

For case 2, the problem with this is that replies will not work with a wildcard URI. If a message is sent over to the gateway, the URI for a return message would include a wildcard, so the receiver of that message would have no idea where to reply to. So if this feature were to be considered, there would need to be something that could throw on the other end to say "Replying doesn't work because you have this setting over here", but I have to think there's a better way, like specifying the proxy address for replies. Based on your experience with K8S, what do you think?

emasdeu:
I think you are right in both cases:

Case1: We got confused because we saw the "info" message in k8s message and thoght it was a error. We will revert this change

Case2: We didn't have the scenrio of the reply, so we did not test it. We are going to try a solution and test the "reply" scenario in k8s.



De: David Boike ***@***.***>
Enviado el: miércoles, 2 de junio de 2021 16:37
Para: Particular/NServiceBus.Gateway ***@***.***>
CC: Masdeu Ferrer, Enric ***@***.***>; Comment ***@***.***>
Asunto: Re: [Particular/NServiceBus.Gateway] Implement a Gateway compatibility with linux and K8s (#267)

Caution, this email may be from a sender outside Wolters Kluwer. Verify the sender and know the content is safe.

Atención, este correo electrónico proviene de un remitente externo a Wolters Kluwer. Proceda a verificarlo para confirmar que el contenido es seguro.

Ok, I get it a little bit more now.

For case 1, I don't think this should be necessary. In this block of the installer<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FParticular%2FNServiceBus.Gateway%2Fblob%2Frelease-3.1%2Fsrc%2FNServiceBus.Gateway%2FInstaller%2FGatewayHttpListenerInstaller.cs%23L40-L51&data=04%7C01%7Cenric.masdeu%40wolterskluwer.com%7Cfd753d8b153241bc126208d925d3f2f6%7C8ac76c91e7f141ffa89c3553b2da2c17%7C0%7C0%7C637582414513311898%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=FCvoUjDcXPqGfPbng6BX6Ey1GdQbIfqmoUEeUgw8gVQ%3D&reserved=0> we're essentially already doing the same check and performing an early exit if you're not on Windows. Do you see something else happening on your end? Is it possible those conditions need to be tweaked?

For case 2, the problem with this is that replies will not work with a wildcard URI. If a message is sent over to the gateway, the URI for a return message would include a wildcard, so the receiver of that message would have no idea where to reply to. So if this feature were to be considered, there would need to be something that could throw on the other end to say "Replying doesn't work because you have this setting over here", but I have to think there's a better way, like specifying the proxy address for replies. Based on your experience with K8S, what do you think?

-

DavidBoike:
Sounds good, let us know how it goes.

alexisadm:
Hi @DavidBoike ,

I'm Alexis and I'm part of the team of architects for the Wolters project. I was one of those who made the change that we are validating with you. As @emasdeu comments, point 1 is a false positive and we interpret it wrong, a point 2 we have not tested the replies and think that they should not work because header
  "NServiceBus.OriginatingSite" ", then we propose:
1 . Rollback point 1 because it doesn't make any sense.
2. Define the option "ChannelURIEnableWildcards".
3. Define in the "Channels" a property called "ProxyAddress" so that if is defined it is used in the header "NServiceBus.OriginatingSite" this will allow the reply to work correctly and point to the cluster proxy.

If that's okay with you, we'll get on with those changes.

Kind regards,
Alexis

DavidBoike:
I would think (but I'd like to get an opinion from either @andreasohlund or @timbussmann as well) that it would be enough to have a defined ProxyAddress. If present, that would be the value used for replies, and then we would run the no-wildcards check on THAT instead of the main address. You wouldn't need any ChannelUriEnableWildcards type property.

alexisadm:
Hi @DavidBoike ,

You are right if we define the "ProxyAddress" property, the wilcards verification could be done on it and the option "ChannelUriEnableWildcards" would not be necessary, we did not want to be so invasive in your code, so I prepare a new PR for your approval?

Kind regards,
Alexis

andreasohlund:
> If present, that would be the value used for replies, and then we would run the no-wildcards check on THAT instead of the main address

sounds good

timbussmann:
agreed. Sounds like a good idea at first glance 👍

DavidBoike:
Replaced by https://github.com/Particular/NServiceBus.Gateway/pull/284

PR #284, opened 2021-06-17
Add ProxyAddress to channel

alexisadm:
Hi all,
As we have talked on this PR (Particular#267), I send a new pull request. target to 3.1  version.

Kind regards,
Alexis

Issue #103, opened 2018-10-03
Only the last registered receive channel is used


andreasohlund:
### Symptoms

When registering more than one receive channel, only the last registered channel is used.

### Who's affected

v3.0.0 Users configuring more than one receive channel

### Details

Originally raised by @welshdave  in https://github.com/Particular/NServiceBus.Gateway/issues/98

PR #285, opened 2021-06-21
Add ProxyAddress

DavidBoike:
From @alexisadm in https://github.com/Particular/NServiceBus.Gateway/pull/284

### DIFF
--- a/src/NServiceBus.Gateway.AcceptanceTests/When_doing_request_reply_with_proxy_address.cs
+++ b/src/NServiceBus.Gateway.AcceptanceTests/When_doing_request_reply_with_proxy_address.cs
@@ -5,29 +5,28 @@
     using Configuration.AdvancedExtensibility;
     using NUnit.Framework;
 
-    public class When_doing_request_reply : NServiceBusAcceptanceTest
+    public class When_doing_request_reply_with_proxy_address: NServiceBusAcceptanceTest
     {
         [Test]
-        public async Task Callback_should_be_fired()
+        public async Task Callback_should_be_fired_when_SideAProxy_IsBehind_Proxy()
         {
             var context = await Scenario.Define<Context>()
-                .WithEndpoint<SiteA>(
-                    b => b.When(async (bus, c) =>
-                    {
-                        var options = new SendOptions();
-                        options.RouteToSites("SiteB");
-                        c.Response = await bus.Request<int>(new MyRequest(), options);
-                        c.GotCallback = true;
-                    }))
-                .WithEndpoint<SiteB>()
-                .Done(c => c.GotCallback)
-                .Run();
+                              .WithEndpoint<SiteA>(
+                                  b => b.When(async (bus, c) =>
+                                      {
+                                          var options = new SendOptions();
+                                          options.RouteToSites("SiteB");
+                                          c.Response = await bus.Request<int>(new MyRequest(), options);
+                                          c.GotCallback = true;
+                                      }))
+                              .WithEndpoint<SiteB>()
+                              .Done(c => c.GotCallback)
+                              .Run();
 
             Assert.IsTrue(context.GotCallback);
             Assert.AreEqual(1, context.Response);
         }
 
-        
         public class Context : ScenarioContext
         {
             public bool GotCallback { get; set; }
@@ -39,14 +38,14 @@
             public SiteA()
             {
                 EndpointSetup<GatewayEndpoint>(c =>
-                {
-                    c.MakeInstanceUniquelyAddressable("1");
-                    c.EnableCallbacks();
+                    {
+                        c.MakeInstanceUniquelyAddressable("1");
+                        c.EnableCallbacks();
 
-                    var gatewaySettings = c.GetSettings().Get<GatewaySettings>();
-                    gatewaySettings.AddReceiveChannel("http://localhost:25697/SiteA/");
-                    gatewaySettings.AddSite("SiteB", "http://localhost:25699/SiteB/");
-                });
+                        var gatewaySettings = c.GetSettings().Get<GatewaySettings>();
+                        gatewaySettings.AddReceiveChannel("http://+:25698/SiteA/", "http", 10, true, "http://localhost:25698/SiteA/");
+                        gatewaySettings.AddSite("SiteB", "http://localhost:25699/SiteB/");
+                    });
             }
         }
 
--- a/src/NServiceBus.Gateway.AcceptanceTests/When_doing_request_reply_with_proxy_address.cs
+++ b/src/NServiceBus.Gateway.AcceptanceTests/When_doing_request_reply_with_proxy_address.cs
@@ -5,7 +5,7 @@
     using Configuration.AdvancedExtensibility;
     using NUnit.Framework;
 
-    public class When_doing_request_reply_with_proxy_address: NServiceBusAcceptanceTest
+    public class When_doing_request_reply_with_proxy_address : NServiceBusAcceptanceTest
     {
         [Test]
         public async Task Callback_should_be_fired_when_SideAProxy_IsBehind_Proxy()
--- a/src/NServiceBus.Gateway.sln.DotSettings
+++ b/src/NServiceBus.Gateway.sln.DotSettings
@@ -592,6 +592,9 @@ II.2.12 &lt;HandlesEvent /&gt;&#xD;
 	<s:String x:Key="/Default/CodeStyle/Naming/XamlNaming/UserRules/=XAML_005FFIELD/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="" Suffix="" Style="AaBb" /&gt;</s:String>
 	<s:String x:Key="/Default/CodeStyle/Naming/XamlNaming/UserRules/=XAML_005FRESOURCE/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="" Suffix="" Style="AaBb" /&gt;</s:String>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpFileLayoutPatternsUpgrade/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpKeepExistingMigration/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpPlaceEmbeddedOnSameLineMigration/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpUseContinuousIndentInsideBracesMigration/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EAddAccessorOwnerDeclarationBracesMigration/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EAlwaysTreatStructAsNotReorderableMigration/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EMigrateBlankLinesAroundFieldToBlankLinesAroundProperty/@EntryIndexedValue">True</s:Boolean>
--- a/src/NServiceBus.Gateway/Channels/Channel.cs
+++ b/src/NServiceBus.Gateway/Channels/Channel.cs
@@ -11,14 +11,15 @@ namespace NServiceBus.Gateway.Channels
         /// The type of the channel.
         /// </summary>
         public string Type { get; set; }
-        
+
         /// <summary>
         /// The address to receive/send on.
         /// </summary>
         public string Address { get; set; }
 
         /// <summary>
-        /// Gets or sets the proxy address using as a public.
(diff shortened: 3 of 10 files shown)


src/NServiceBus.Gateway.sln.DotSettings
```diff
@@ -592,6 +592,9 @@ II.2.12 &lt;HandlesEvent /&gt;&#xD;
...
 	<s:String x:Key="/Default/CodeStyle/Naming/XamlNaming/UserRules/=XAML_005FRESOURCE/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="" Suffix="" Style="AaBb" /&gt;</s:String>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpFileLayoutPatternsUpgrade/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpKeepExistingMigration/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpPlaceEmbeddedOnSameLineMigration/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpUseContinuousIndentInsideBracesMigration/@EntryIndexedValue">True</s:Boolean>
```
DavidBoike, line 597: @timbussmann I wouldn't worry about these. The file has been removed in master.

DavidBoike:
Hang on, I may have overlooked something.

@alexisadm did you see this section of our documentation? I just now found it for the first time:

https://docs.particular.net/nservicebus/gateway/multi-site-deployments?version=gateway_4#incoming-channels-with-wildcard-uris

It seems to indicate that we don't need to create a ProxyAddress concept because the preferred (and already documented) method is to add another channel with `isDefault: true` and then that essentially **is** the proxy address, with no software changes required.

Can you confirm whether or not that works for you?

If so then we don't need either of these PRs. Also, if that's the case, I'd like to apologize to you for putting so much time into this when there was a fairly simple solution if we just could have pointed you to it. Several of us looked at it and we all missed it!

If it turns out that doesn't work, let us know and then we can continue with this PR.

alexisadm:
> Hang on, I may have overlooked something.
> 
> @alexisadm did you see this section of our documentation? I just now found it for the first time:
> 
> https://docs.particular.net/nservicebus/gateway/multi-site-deployments?version=gateway_4#incoming-channels-with-wildcard-uris
> 
> It seems to indicate that we don't need to create a ProxyAddress concept because the preferred (and already documented) method is to add another channel with `isDefault: true` and then that essentially **is** the proxy address, with no software changes required.
> 
> Can you confirm whether or not that works for you?
> 
> If so then we don't need either of these PRs. Also, if that's the case, I'd like to apologize to you for putting so much time into this when there was a fairly simple solution if we just could have pointed you to it. Several of us looked at it and we all missed it!
> 
> If it turns out that doesn't work, let us know and then we can continue with this PR.

Hi @DavidBoike ,

We've tested your's approach but does not work, got next errors (hide valid address)

nhandled exception. System.Exception: Failed to start listener for http://validproxy-k8s.address:60396/MacGateway/ make sure that you have admin privileges
 ---> System.Net.HttpListenerException (99): Address not available
   at System.Net.HttpEndPointManager.GetEPListener(String host, Int32 port, HttpListener listener, Boolean secure)
   at System.Net.HttpEndPointManager.AddPrefixInternal(String p, HttpListener listener)
   at System.Net.HttpEndPointManager.AddListener(HttpListener listener)
   at System.Net.HttpListener.Start()
   at NServiceBus.Gateway.Channels.Http.HttpChannelReceiver.Start(String address, Int32 maxConcurrency, Func`2 dataReceivedOnChannel)
   --- End of inner exception stack trace ---
   at NServiceBus.Gateway.Channels.Http.HttpChannelReceiver.Start(String address, Int32 maxConcurrency, Func`2 dataReceivedOnChannel)
   at NServiceBus.Gateway.Receiving.SingleCallChannelReceiver.Start(Channel channel, Int32 maxConcurrency, Func`2 receivedHandler)
   at NServiceBus.Features.Gateway.GatewayReceiverStartupTask.OnStart(IMessageSession context)
   at NServiceBus.Features.FeatureStartupTask.PerformStartup(IMessageSession session) in /_/src/NServiceBus.Core/Features/FeatureStartupTask.cs:line 25
   at NServiceBus.Features.FeatureStartupTaskController.Start(IBuilder builder, IMessageSession messageSession) in /_/src/NServiceBus.Core/Features/FeatureStartupTaskController.cs:line 21
   at NServiceBus.Features.FeatureActivator.StartFeatures(IBuilder builder, IMessageSession session) in /_/src/NServiceBus.Core/Features/FeatureActivator.cs:line 68
   at NServiceBus.StartableEndpoint.Start() in /_/src/NServiceBus.Core/StartableEndpoint.cs:line 64
   at NServiceBus.HostingComponent.Start(IStartableEndpoint startableEndpoint) in /_/src/NServiceBus.Core/Hosting/HostingComponent.cs:line 76
   at NServiceBus.ExternallyManagedContainerHost.Start(IBuilder externalBuilder) in /_/src/NServiceBus.Core/Hosting/ExternallyManagedContainerHost.cs:line 49
   at TASuite.Commons.Infrastructure.NServiceBusCore.HostingExtensions.StartWithExternalServiceProviderAsync(IStartableEndpointWithExternallyManagedContainer instance, IServiceProvider provider)
   at TASuite.Commons.Infrastructure.NServiceBusCore.StandaloneHost.StartAsync(CancellationToken cancellationToken)
   at Microsoft.Extensions.Hosting.Internal.Host.StartAsync(CancellationToken cancellationToken)
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.RunAsync(IHost host, CancellationToken token)
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.RunAsync(IHost host, CancellationToken token)
   at TASuite.MAC.CDA.Sender.Program.<Main>(String[] args)
Aborted (core dumped)

with configuration :

  "Gateway": {
    "SiteName": "MacSite",
    "SecurityProtocolType": "Tls12",
    "ActivateK8SCompatibility": "true",
    "Channels": [
      {
        "Address": "http://*:60395/MacGateway/",
        "ChannelType": "Http",
        "MaxConcurrency": "10",
        "IsDefault": false
      },
      {
        "Address": "http://validproxy-k8s.address:60396/MacGateway/",
        "ChannelType": "Http",
        "MaxConcurrency": "10",
        "IsDefault": true
      }

We thinks that Pull Request should go on.

Kind regards,
Alexis

src/NServiceBus.Gateway.AcceptanceTests/When_doing_request_reply_with_proxy_address.cs
```diff
@@ -0,0 +1,78 @@
...
+{
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using Configuration.AdvancedExtensibility;
+    using NUnit.Framework;
+
+    public class When_doing_request_reply_with_proxy_address : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Callback_should_be_fired_when_SideAProxy_IsBehind_Proxy()
```
timbussmann, line 11:
```suggestion
        public async Task Callback_should_be_fired_when_SiteAProxy_IsBehind_Proxy()
```

src/NServiceBus.Gateway/Settings/GatewaySettings.cs
```diff
@@ -103,14 +103,13 @@ public void AddSite(string siteKey, string address, string type = "http", bool l
...
-        /// Adds a receive channel that the gateway should listen to.
-        /// </summary>
+        /// <summary>Adds a receive channel that the gateway should listen to.</summary>
         /// <param name="address">The channel address.</param>
-        /// <param name="maxConcurrency">Maximum number of receive connections. Default is `1`.</param>
         /// <param name="type">The channel type. Default is `http`.</param>
+        /// <param name="maxConcurrency">Maximum number of receive connections. Default is `1`.</param>
         /// <param name="isDefault">True if this should be the default channel for send operations. Default is `false`.</param>
-        public void AddReceiveChannel(string address, string type = "http", int maxConcurrency = 1, bool isDefault = false)
+        /// <param name="proxyAddress">The proxy Address to use when gateway is behind proxy, is used in reply on headers.</param>
```
timbussmann, line 111: > is used in reply on headers

that part is a bit confusing to read imo

timbussmann (approved):

timbussmann:
> https://docs.particular.net/nservicebus/gateway/multi-site-deployments?version=gateway_4#incoming-channels-with-wildcard-uris

does indeed sound like it should do what this PR also tries to attempt.

> System.Net.HttpListenerException (99): Address not available

that sounds like a different issue that's not directly related to the actual feature? I think we should further investigate why such an error would occur rather than just building the same feature twice?

DavidBoike:
Hmm, I think the original feature may have been poorly thought out. Bat signal to @timbussmann & @andreasohlund:

For instance, this works, which is taken straight from the [acceptance test](https://github.com/Particular/NServiceBus.Gateway/blob/master/src/NServiceBus.Gateway.AcceptanceTests/When_listening_on_wildcard_uri.cs), which unfortunately only asserts that this configuration won't throw, not that it does anything useful IRL:

```csharp
gateway.AddReceiveChannel("http://+:25701/WildcardB/");
gateway.AddReceiveChannel("http://localhost:25700/WildcardB/", isDefault: true);
```

But with that configuration `localhost` would be used in the reply header. That would not work IRL. If all others can reply to "localhost" then why bother with a wildcard at all? I honestly don't see how the solution enables the case mentioned in the [2.0.1 release notes](https://github.com/Particular/NServiceBus.Gateway/releases/2.0.1) in the first place. My only theory is that it has something to do with this later bug that was fixed [Only the last registered receive channel is used](https://github.com/Particular/NServiceBus.Gateway/issues/103). Perhaps prior to that "fix" you could have one URL for the channel and the second URL for the proxy and it only listened on the wildcard and everything was cool.

For every receive channel, it's going to try to bind to that address. But I think the situation here especially with K8S is that you will never be able to bind to the publicly-addressable URL.

So in order to "make this work" I'm considering the different things that could happen with multiple receive channels and they both seem bad:

* Don't try to bind to an address if it's marked as a default channel—what about the people who expect the binding?
* Try to bind to an address if it's marked as a default channel, but catch and ignore any bind exceptions as long as there's a different channel. But then, what happens if you *do* want to bind to that address? You just get a warning that "Sorry, couldn't bind to that address, but you may have meant it as a proxy address so I'll just fail silently" ?

Seems like "add another channel" is the incorrect response to this problem in the first place. The correct answer is "we have one channel, but it has a different internal/external address" which seems exactly what this PR aims to solve.

Do you agree?

DavidBoike:
Talked internally, and we agree that the feature may have worked in the past, and that it was probably "broken" later by a "bufix" to ensure all regsistered channels were bound.

But the proxy address as we have it implemented here is funny, because the value only matters for the isDefault channel. You could specify the rest but they'd never get used for anything. It also feels wrong to declare a publicly-addressable address by adding a "new channel".

So internally we have agreed to implement the following:

```
// All channels get bound
gateway.AddReceiveChannel("http://+:25701/Whatever/");

// New API
gateway.SetReplyToAddress("http", "http://mydomain.whatever.com/whatever");
```

It's of course up to you to make sure that the load balancer routes the reply-to address to a valid receive channel.

This serves the same purpose as the proxy address but is more intention-revealing. We'd also check to make certain that at least one channel of the same type exists. It wouldn't do to have an HTTPS reply to address but no HTTPS bindings. (Let alone FTP or anything else weird.)

And then we'd also only make sure that the reply-to address didn't have wildcards. If it was provided we wouldn't do a wildcard check on individual channel addresses.

@alexisadm does this sound good to you? All I want is a yes or no. I'll implement this tomorrow—I've already sent you down enough rabbit holes, I don't want to ask you to submit another PR.

alexisadm:
Hi all,

Hi all, The approach described on NServiceBus documentation didn't solve
our´s issues because it create two channels and default has not a valid
address inside pod.

We have two issues :
1. On startup application check than default channel have not wildcard. On
Kubernetes that is the only way to register listener on any port because we
test localhost and does'n work.
2. If define a wildcard the reply will not work.

Maybe, all time we are talking about coherency , because "ProxyAddress" is
not a good name, but naming it as "ReplyAddress" or whatever that alway be
associated to de channel because the wildcard control will be over that
property and in message headers for a reply, will a good solution for us.
In addition to that , in IaaS solutions with disaster recovering solution
and different network interfaces you can have different load balancers on
different ports, set ReplyAddress globally will not solve that issue, that
is why ReplyAddress need to be set on channel definition, maybe this is a
"YAGNI" but is a collection of channels.

Kind regards,
Alexis

DavidBoike:
@alexisadm I have raised https://github.com/Particular/NServiceBus.Gateway/pull/287 against master as the alternative to create a `SetReplyToAddress(string address, string type = "http")` API.

If @timbussmann and/or @andreasohlund approve that, I'll integrate it with the current release branch and do a new release for you.

src/NServiceBus.Gateway/Channels/Channel.cs
```diff
@@ -28,6 +34,14 @@ internal static Channel Parse(string s)
             };
         }
 
+        /// <summary>
```
andreasohlund, line 37: Nitpick: internals doesn't need xml docs

src/NServiceBus.Gateway/Settings/GatewaySettings.cs
andreasohlund: How about

>The address to which replies should be sent should the default channel not be publicly available, eg. when running behind a proxy

andreasohlund: This makes me wonder if the name ProxyAddress is too tied to HTTP? Would it be relevant to other channel types like FTP?

I think this is just another sign of the design with custom channels isn't really thought out and we just have to live with it for now

andreasohlund (approved):

alexisadm:
Hi All , I did changes suggested by @andreasohlund. We have a local version in production, what are next steps?

DavidBoike:
Closing this PR. SetReplyToUri will be implemented instead in https://github.com/Particular/NServiceBus.Gateway/pull/287 and then backported to release-3.1 in https://github.com/Particular/NServiceBus.Gateway/pull/288.

@alexisadm Having a ProxyAddress (even named anything else) just doesn't make sense because when it comes right down to it, a Gateway only has one ReplyTo concept it can put on outgoing communications. Messages that come off the wire from HTTP get turned into a message, and then they're processed, and if you do a Reply the context of which channel it came in off of is lost. So being able to set an individual ReplyTo address for a channel that isn't the default channel is a setting that would do nothing…except cause confusion. Configuring one ReplyTo thing reflects the reality of what's going on.

PR #287, opened 2021-06-24
Add SetReplyToUri() API

DavidBoike:


src/NServiceBus.Gateway/Settings/GatewaySettings.cs
```diff
@@ -123,6 +123,21 @@ public void AddReceiveChannel(string address, string type = "http", int maxConcu
...
         }
 
+        /// <summary>
+        /// Sets the reply-to address for messages sent from this Gateway. Useful for setting a publicly-accessible
+        /// load balancer address that will be routed to this Gateway, but the local system would be unable
+        /// to bind an HTTP listener (or other type of listener) to that address.
+        /// </summary>
+        /// <param name="address">The publicly-accessible address to include as the Reply-To address.</param>
+        /// <param name="type">The address type. Default is `http`. Must match one of the incoming receive channels.</param>
+        public void SetReplyToAddress(string address, string type = "http")
```
timbussmann, line 133: what other types would be available/relevant here? http/https can already be derived from the address itself as far as I can tell from the examples given?

also, seems like this would allow passing "https://..." in the address although the type remains at "http" as a default?

timbussmann: `Address` feels a bit ambiguous especially in the NServiceBus context? Would something like `ReplyToUri` `ReplyToChannel` or something like that be more clear?

src/NServiceBus.Gateway/Settings/GatewaySettings.cs
```diff
@@ -149,6 +164,11 @@ internal static List<ReceiveChannel> GetConfiguredChannels(ReadOnlySettings sett
             return settings.TryGet(out List<ReceiveChannel> channels) ? channels : new List<ReceiveChannel>();
         }
 
+        internal static (string type, string address) GetReplyToAddress(ReadOnlySettings settings)
```
timbussmann, line 167: will this be backported? Using tuples might be an issue when targeting the latest major release?

src/NServiceBus.Gateway/Settings/GatewaySettings.cs
DavidBoike: Most of the time, channel types are http and https, but it's really whatever you choose to create. You could create ftp. You could create "fred" if you wanted, and that could actually communicate via http, if you created a class to do that. Look at how with our built-in code, [channel type attributes decorate the class](https://github.com/Particular/NServiceBus.Gateway/blob/master/src/NServiceBus.Gateway/Channels/Http/HttpChannelSender.cs#L12-L14) with http and https to provide the built-in implementations.

So there's really no need for the "channel type" and URI scheme to even agree. That's just the way the Gateway is.

For the method name, I agree that `ReplyToUri` would be better as that doesn't presume anything. I'll change it to that.

src/NServiceBus.Gateway/Settings/GatewaySettings.cs
DavidBoike: Yes it will be backported to the current release branch, but not further than that. If it's an issue there, I'll switch to an explicit`Tuple<string, string>` or create an internal DTO or whatever works.

DavidBoike:
PR to backport to release-3.1 here: https://github.com/Particular/NServiceBus.Gateway/pull/288

timbussmann (approved):

PR #293, opened 2021-08-10
GitHubSync update - master

particularbot6:
This is an automated synchronization PR.

The following source template repositories were used:
* particular/RepoStandards


andreasohlund (approved):

src/NServiceBus.Gateway/NServiceBus.Gateway.csproj
```diff
@@ -11,6 +11,14 @@
...
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Remove="obsoletes-v4.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Include="obsoletes-v4.cs" />
+  </ItemGroup>
+
```
adamralph, line 21: Did VS do something weird here? I think we can remove these lines?

```suggestion
```

src/NServiceBus.Gateway/obsoletes-v4.cs
```diff
@@ -0,0 +1,21 @@
...
+    [ObsoleteEx(
+        TreatAsErrorFromVersion = "8.0.0",
+        RemoveInVersion = "9.0.0",
+        ReplacementTypeOrMember = nameof(DataReceivedOnChannelEventArgs))]
+    public class DataReceivedOnChannelArgs : EventArgs
+    {
+        public IDictionary<string, string> Headers { get; set; }
+
+        public Stream Data { get; set; }
+    }
```
adamralph, line 18: The version numbers are wrong, and I think we can remove the implementation?

Also, how is it that we don't need to suppress IDE1006 here?

```suggestion
    [ObsoleteEx(
        TreatAsErrorFromVersion = "4.0.0",
        RemoveInVersion = "5.0.0",
        ReplacementTypeOrMember = nameof(DataReceivedOnChannelEventArgs))]
    public class DataReceivedOnChannelArgs { }
```

adamralph (changes requested):

src/NServiceBus.Gateway/obsoletes-v4.cs
adamralph: > Also, how is it that we don't need to suppress IDE1006 here?

I think it's because of https://github.com/Particular/NServiceBus.Gateway/pull/293#discussion_r688537251 - I think those lines are removing this file from compilation, which is definitely not what we want.

src/NServiceBus.Gateway/NServiceBus.Gateway.csproj
andreasohlund: Yea, fixed

src/NServiceBus.Gateway/obsoletes-v4.cs
andreasohlund: duh, fixed now!

andreasohlund: and it make the compile error show up like you said

adamralph (approved):

PR #334, opened 2021-11-23
Stop using the static Current property to prevent cleanup operations to execute on the wrong endpoint

mauroservienti:
I noticed while working on the [RavenDB implementation](https://github.com/Particular/NServiceBus.Gateway.RavenDB), that when a test is composed of two or more endpoints the cleanup phase is run against the `GatewayTestSuiteConstraints` instance that has been created last.

The last endpoint that is started replaces the `Current` static that is then used in all cleanup phases.

### DIFF
--- a/src/NServiceBus.Gateway.AcceptanceTests/GatewayEndpoint.cs
+++ b/src/NServiceBus.Gateway.AcceptanceTests/GatewayEndpoint.cs
@@ -13,7 +13,8 @@
             {
                 var endpointName = endpointCustomizationConfiguration.CustomEndpointName ?? configuration.GetSettings().EndpointName();
 
-                var deduplicationConfiguration = GatewayTestSuiteConstraints.Current.ConfigureDeduplicationStorage(
+                var constraints = new GatewayTestSuiteConstraints();
+                var deduplicationConfiguration = constraints.ConfigureDeduplicationStorage(
                     endpointName,
                     configuration,
                     runDescriptor.Settings)
@@ -23,7 +24,7 @@
 
                 configuration.GetSettings().Set(gatewaySettings);
 
-                runDescriptor.OnTestCompleted(_ => GatewayTestSuiteConstraints.Current.Cleanup());
+                runDescriptor.OnTestCompleted(_ => constraints.Cleanup());
 
                 configurationBuilderCustomization(configuration);
             });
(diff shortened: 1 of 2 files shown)


andreasohlund (approved):

src/NServiceBus.Gateway.AcceptanceTests/GatewayTestSuiteConstraints.cs
```diff
@@ -2,6 +2,5 @@
 {
     public partial class GatewayTestSuiteConstraints : IGatewayTestSuiteConstraints
     {
-        public static GatewayTestSuiteConstraints Current = new GatewayTestSuiteConstraints();
```
andreasohlund, line 5: Will this break the downstreams?

andreasohlund:
@timbussmann since we're updating the downstreams anyway we could make sure to include this change as well?

src/NServiceBus.Gateway.AcceptanceTests/GatewayEndpoint.cs
```diff
@@ -13,7 +13,8 @@ public override Task<EndpointConfiguration> GetConfiguration(RunDescriptor runDe
             {
                 var endpointName = endpointCustomizationConfiguration.CustomEndpointName ?? configuration.GetSettings().EndpointName();
 
-                var deduplicationConfiguration = GatewayTestSuiteConstraints.Current.ConfigureDeduplicationStorage(
+                var constraints = new GatewayTestSuiteConstraints();
```
timbussmann, line 16: would it make sense to just snapshot the instance here?

```suggestion
                var constraints = GatewayTestSuiteConstraints.Current;
```
I'm not sure about the usage of the current static property, but IIRC in core, this can also be used to replace default settings when needed on a specific test? not sure I remember the details correctly though.

src/NServiceBus.Gateway.AcceptanceTests/GatewayTestSuiteConstraints.cs
timbussmann: In core it gives access to special properties about whether the configuration supports certain features or not, but that doesn't seem to be the case with the gateway

andreasohlund:
Talked to @timbussmann and he's fine with the changes, merging.

@mauroservienti we'll make sure to update the downstreams

PR #335, opened 2021-11-23
Update to core 8.0.0-alpha.1900

andreasohlund:


andreasohlund:
@timbussmann changes we talked about are in now

src/NServiceBus.Gateway/Gateway.cs
```diff
@@ -54,29 +54,30 @@ protected override void Setup(FeatureConfigurationContext context)
...
+            var logicalGatewayAddress = new QueueAddress(context.Settings.EndpointQueueName(), null, null, "gateway");
             var replyToAddress = GetReplyToAddress(context.Settings, channelManager);
 
-            var sender = new GatewayMessageSender(
-                gatewayInputAddress,
+            context.Services.AddSingleton(b => new GatewayMessageSender(
+                b.GetService<ITransportAddressResolver>().ToTransportAddress(logicalGatewayAddress),
                 new MessageNotifier(),
-                context.Settings.LocalAddress(),
+                b.GetService<ReceiveAddresses>().MainReceiveAddress,
```
timbussmann, line 65: can the gateway somehow be run in send-only mode that would fail to resolve the service?

src/NServiceBus.Gateway/Gateway.cs
```diff
@@ -54,29 +54,30 @@ protected override void Setup(FeatureConfigurationContext context)
...
-
             var retryPolicy = context.Settings.Get<Func<IncomingMessage, Exception, int, TimeSpan>>("Gateway.Retries.RetryPolicy");
 
+            var logicalGatewayAddress = new QueueAddress(context.Settings.EndpointQueueName(), null, null, "gateway");
             var replyToAddress = GetReplyToAddress(context.Settings, channelManager);
 
-            var sender = new GatewayMessageSender(
-                gatewayInputAddress,
+            context.Services.AddSingleton(b => new GatewayMessageSender(
+                b.GetService<ITransportAddressResolver>().ToTransportAddress(logicalGatewayAddress),
```
timbussmann, line 63: shouldn't this use GetRequiredService (all the resolves here)?

andreasohlund:
Pushed a fix to use GetRequiredService where needed

src/NServiceBus.Gateway/Gateway.cs
andreasohlund: Not sure but since the gateway is using satellites to do the forwarding you can't use it if the endpoint is sendonly?

Unless there is an option to run satellites but still be send only?

andreasohlund: And the previous call to context.Settings.LocalAddress() would have thrown on a sendonly endpoint

timbussmann: > Unless there is an option to run satellites but still be send only?

I don't think there is an option but I think this fails silently

> And the previous call to context.Settings.LocalAddress() would have thrown on a sendonly endpoint

I don't think that's true? https://github.com/Particular/NServiceBus/blob/7.5.0/src/NServiceBus.Core/SettingsExtensions.cs#L57 does not check for send-only?

andreasohlund: So the exception isn't correct then? https://github.com/Particular/NServiceBus/blob/7.5.0/src/NServiceBus.Core/SettingsExtensions.cs#L54

timbussmann (approved):

PR #386, opened 2022-05-11
Remove netcoreapp3.1

DavidBoike:


src/NServiceBus.Gateway/Channels/Http/HttpChannelSender.cs
```diff
@@ -4,49 +4,55 @@ namespace NServiceBus.Gateway.Channels.Http
...
+                        throw new Exception("Invalid header");
+                    }
                 }
             }
-            catch (WebException ex)
+
+            HttpStatusCode statusCode;
+
+            //todo make the receiver send the md5 back so that we can double check that the transmission went ok
+            using (var httpClient = new HttpClient(httpClientHandler, disposeHandler: false))
```
danielmarbach, line 50: Is this a good idea? I get that you are making sure the handler doesn't get disposed. Yet almost everwhere you see guidance to not dispose the http client (because of the handler). Still wouldn't it be better since we are relying on service collection now to take a dependency on Microsoft.Extensions.Http stuff and then rely on that integration to get for example a named client that users can control and override if necessary? That would also allow them to plug in polly extensions...

bording (approved):

src/NServiceBus.Gateway/Channels/Http/HttpChannelSender.cs
danielmarbach: While you are right the culprit is mostly the handler there is a ton of guidance around HTTP Client that clearly identifies disposing HTTP Client as an anti-pattern. Just a few examples of posts around this topic

https://www.thecodebuzz.com/using-httpclient-best-practices-and-anti-patterns/
https://www.rahulpnath.com/blog/are-you-using-httpclient-in-the-right-way/
https://www.stevejgordon.co.uk/httpclient-creation-and-disposal-internals-should-i-dispose-of-httpclient

> [HttpClient](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=net-6.0) is intended to be instantiated once and re-used throughout the life of an application. Instantiating an HttpClient class for every request will exhaust the number of sockets available under heavy loads. This will result in SocketException errors.

https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient

So the minimal change would then have been to make the HTTP client static and not introduce a very uncommon pattern of using the HTTP Client. Can we somehow follow up on this change? I'm happy to help

PR #413, opened 2022-09-01
Make gateway aware of databus contenttype header 

ramonsmits:
Makes gateway aware of databus contenttype header which had the same `NServiceBus.DataBus.` namespace prefix which clashes with the BLOB id's.


ramonsmits:
Obsolete due to:

- https://github.com/Particular/NServiceBus/pull/6531
- https://github.com/Particular/NServiceBus/pull/6527

